PR #3, opened 2024-12-19
Update state.py

kdotom:
Updating line which gets "num_devices" to get the backend in general (vs "gpu" specifically)

This is more of a question/confirmation than a necessary PR. I made this general to have it work with a cpu, but feel free to leave it as is if there's a deeper dependency.


ymahlau:
Thanks for catching this! The backend should definitely not be hardcoded there

PR #13, opened 2025-02-04
Implementation of Periodic Boundary Condition and Bugfix for plotting on Mac

txdai:
Implemented periodic boundary condition that wraps around the boundaries. Can be applies to any pair of xyz planes.

Bug fix for plotting on Mac that has DPI scaling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ scripts/configs/run_config.yaml
 *.jsonl
 .aider*
 .DS_Store
+.cursorignore
 site
 
 # Byte-compiled / optimized / DLL files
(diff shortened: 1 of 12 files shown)


ymahlau:
Thanks for implementing this feature! I will merge this PR now.

ymahlau:
As a minor comment, I believe that this implementation is a bit overcomplicated. The curl operation already natively supported the periodic boundary if the padding size would be set to zero at the corresponding side. It might be worthwhile to simplify the implemetation in a future refactoring

txdai:
> As a minor comment, I believe that this implementation is a bit overcomplicated. The curl operation already natively supported the periodic boundary if the padding size would be set to zero at the corresponding side. It might be worthwhile to simplify the implemetation in a future refactoring

I am planning on adding a Fourier Detector/Diffractive Detector. This will be fixed in the next PR.

PR #16, opened 2025-02-20
Implmentations for various features regarding photonic devices

txdai:
When optimizing for metagratings, a measurement for the diffractive properties would be necessary, and to check over multiple frequencies, a simulation with broadband (gaussian pulse) source is also favorable. During optimization, using gray scale devices in addition to binary devices would also improve stability. Hence the new features here.

The gaussian pulse source seems to be somewhat buggy at the moment but I cannot identify the errors. The other two functionalities seems to work fine.

### DIFF
--- a/docs/tutorials/constraint_mapping.md
+++ b/docs/tutorials/constraint_mapping.md
@@ -3,14 +3,14 @@
 When using inverse design with FDTDX, fabrication constraints have to be specified. The basic building block for an object optimizable by inverse design is a Device:
 
 ```python
-from fdtdx.objects.multi_material.device import Device
+from fdtdx.objects.multi_material.device import DiscreteDevice
 from fdtdx.core.physics import constants
 
 permittivity_config = {
     "Air": constants.relative_permittivity_air,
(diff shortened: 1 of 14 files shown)


ymahlau:
Hi, could you maybe split this PR into multiple PR for the different features? It is very difficult to work through such a large PR, because so many things change and many bugs may arise at many places.

Some comments regarding the features:
- I dislike the distinction between discrete and continous device. The only difference there is the constraint mapping and that can be set as an argument. I think a better way to do this would be to only have a Device. The we can make a class called ContinousDevice, which automatically adds the continous transition to the module list. This would prevent code duplication between a discrete and continous device.
- is the ContinuousPermittivityTransition supposed to do subpixel smoothing? I do not think this is achieved in the implementation, because every voxel is only updated based on the own value.
- I like the implementation of a TemporalProfile. I think there are some features missing though, for example when the start_time of a source is set to something other than 0, the profile is activated at the wrong time. 
- pre-commit is complaining about some unused variables and formatting in the PR

PR #17, opened 2025-02-20
Temporal profile

txdai:
This pull request focuses on the temporal profiles of sources. The changes include the introduction of new classes for temporal profiles and modifications to existing methods to utilize these profiles.

Key changes include:

### Introduction of Temporal Profiles:
* Added a new abstract base class `TemporalProfile` and its implementations `SingleFrequencyProfile` and `GaussianPulseProfile` to define how source amplitude varies in time. (`src/fdtdx/objects/sources/source.py`)

### Modifications to Source Class:
* Updated the `Source` class to include a `temporal_profile` attribute, which defines the time-dependent amplitude variation of the source. (`src/fdtdx/objects/sources/source.py`) [[1]](diffhunk://#diff-3d843c188f48ba9e4403e6a60cf646b11a6cebf81f58e3e5fe5fa4041e33d768R103) [[2]](diffhunk://#diff-3d843c188f48ba9e4403e6a60cf646b11a6cebf81f58e3e5fe5fa4041e33d768R117)

### Refactoring of Update Methods:
* Refactored the `update_E` method to use the `temporal_profile` for calculating the temporal amplitude of the H field. (`src/fdtdx/objects/sources/plane_source.py`)
* Refactored the `update_H` method to use the `temporal_profile` for calculating the temporal amplitude of the E field. (`src/fdtdx/objects/sources/plane_source.py`)

### DIFF
--- a/src/fdtdx/objects/sources/plane_source.py
+++ b/src/fdtdx/objects/sources/plane_source.py
@@ -396,11 +396,9 @@ class PlaneSource(DirectionalPlaneSourceBase, ABC):
         inv_permittivity_slice = inv_permittivities[*self.grid_slice]
 
         # Calculate time points for E and H fields
-        time_E = (time_step - 0.5 + self._time_offset_E[self.vertical_axis]) * delta_t
         time_H = (time_step - 0.5 + self._time_offset_H[self.horizontal_axis]) * delta_t
 
         # Get temporal amplitudes from profile
-        amplitude_E = self.temporal_profile.get_amplitude(time_E, self.period, self.phase_shift)
         amplitude_H = self.temporal_profile.get_amplitude(time_H, self.period, self.phase_shift)
 
         # vertical incident wave part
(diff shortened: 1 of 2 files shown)


ymahlau:
Hi, you mentioned that you encountered some problems with this implementation. Are these issues resolved in the PR? In my small tests, the gaussian temporal envelope seemed to work perfectly fine.

I think it would be good to have the start offset to avoid discontinuities at t=0 as a parameter that can be adjusted by the user. 6 sigma might be a bit excessive for some applications, though it is a good default parameter.

txdai:
I didn't change that though. Once I observed that the fields are not propagating but that might be an error in how I used this source

ymahlau:
Did you simulate for a long enough time? To me it looks like its working. It takes a long time until the pulse appears because you have a lot of standard deviations at the beginning to prevent the discontinouity.

Flux Detector_poynting_flux_0

https://github.com/user-attachments/assets/f3af196a-9e8a-45c5-aa08-c723be5c2cc0

txdai:
I understand, it is working now!

PR #25, opened 2025-03-17
Change tidy3d mode solver import

brandmud:
Hey, thanks for your great work on fdtdx.

Some colleagues and I tried to run the examples provided and ran into an error that occurs when trying to import some functions from tidy3d. It seems like the compute_modes and _compute_modes functions do not exist in the plugins module, but rather in the components module. This was also confirmed by the documentation.

I have simply replaced this import command and now the examples work for us.

Thanks again!


ymahlau:
Thanks for catching this bug! It seems that in tidy3d version 2.8.0 the import path changed. Could you also change the minimum version of tidy3d in the [pyproject.toml](https://github.com/ymahlau/fdtdx/blob/6bd0d89f0d4580d869145ece3a5bf04fb0d1e5d6/pyproject.toml#L21) to 2.8.0? Then I will merge the PR afterwards.

brandmud:
Done! Thanks for the swift reply!

PR #26, opened 2025-03-19
Add precision to _compute_mode in tidy3d

brandmud:
It seems like the new tidy3d version brought along a change in the _compute_modes function, which now expects the precision to be passed explicitly. I have added this in both instances where this method is called.


ymahlau:
Yes, i actually noticed this also. In the development branch it is already changed, but I forgot to do it in the main branch :) Thanks for catching this!

Issue #37, opened 2025-03-31
Detector plot of slice instead of mean


ymahlau:
Currently, the plotting of a 3d volume is performed by simply using the mean pixel value over the corresponding view. However, it would be desirable to offer multiple aggregation functions, especially for plotting a single slice through the volume.

This feature requires the implementation of an indexing system, which calculates the position of the corresponding slice. This indexing should occur using the object coordinate system similar to the constraint system, i.e. -1 for lower boundary, 0 for center, 1 for top boundary

msimicphysics:
I have implemented the slicing option for the 3D volume without aggregation in the EnergyDetector class, providing x, y, and z slices.

msimicphysics:
Sure! I opened a pull request with my current implementation of the energy detector class.

PR #42, opened 2025-03-31
[EnergyDetectorSliceVideo] Added slicing option for energy detector

msimicphysics:
Added slicing capability for the energy detector class. Requires x, y, and z for slicing in the yz, xz, and xy planes, respectively.

### DIFF
--- a/src/fdtdx/objects/detectors/energy.py
+++ b/src/fdtdx/objects/detectors/energy.py
@@ -70,9 +70,8 @@ class EnergyDetector(Detector):
         arr_idx = self._time_step_to_arr_idx[time_step]
 
         if self.as_slices:
-            use_mean = (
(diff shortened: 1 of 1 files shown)


ymahlau:
This looks like a good implementation! Before I merge this PR, could you add an option that keeps the old implementation with mean values? I think the default behavior should be the new implementation with slices, but I had some applications in the past where the mean value was very useful (Thats why it was implemented this way before).

ymahlau:
Thanks! Pre-commit now fails a formatting check, because one of the changed lines violates the formatting standards we use for this project. You can see the problematic line in the failed check log. To fix this, you can either copy the corrected lines from the logs or run "pre-commit run --all" on your local machine to fix the formatting automatically.

ymahlau:
resolves #37

Issue #50, opened 2025-04-28
Mode Decomposition


ymahlau:
For many applications it is necessary to decompose the measured fields into specific modes, e.g. S-parameter calculation, mode crosstalk, etc. Therefore, we have to implement mode decomposition similar to meep:

https://meep.readthedocs.io/en/latest/Mode_Decomposition/

I believe the best way to implement this functionality is not as a seperate function, which takes as input the fields measured by the phasor detector. For best usability, we can also implement a ModeMonitor detector class, which is a subclass of the phasor detector and offers functionality to decompose the fields into specific modes.

msimicphysics:
Hi! I'm currently working on an implementation for this issue. Could you please assign it to me?

PR #63, opened 2025-05-21
[ModeFiltering] - Added ModeOverlap Detector class

msimicphysics:
Added ModeOverlapDetector class

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
         args: [--assume-in-merge]
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.8.2
+    rev: v0.11.10
     hooks:
       # Run the linter.
       - id: ruff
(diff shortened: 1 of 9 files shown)


ymahlau:
Nice Implementation, thanks a lot! 

To avoid code duplication, a future refactoring should try to make the mode detector a subclass of the phasor detector or put the phasor update in a separate function used by both detectors. I will take a look at how to do that best :)

Also, for efficiency reasons, it is probably better to store the mode_E and mode_H fields as a private attribute of the detector instead of in the detector state. I will probably refactor that in the future.

ymahlau:
closes #50

msimicphysics:
> Nice Implementation, thanks a lot!
> 
> To avoid code duplication, a future refactoring should try to make the mode detector a subclass of the phasor detector or put the phasor update in a separate function used by both detectors. I will take a look at how to do that best :)
> 
> Also, for efficiency reasons, it is probably better to store the mode_E and mode_H fields as a private attribute of the detector instead of in the detector state. I will probably refactor that in the future.

Thanks, Yannik, also for your helpful comments! I would appreciate it if you could refactor and implement your suggestions in a future release.

Issue #88, opened 2025-06-27
PeriodicBoundaryState, BaseBoundaryState, BoundaryState


ymahlau:
Rename and restructure the boundary objects and their states. Currently, many typing rules are violated.

ymahlau:
Also add typing to specify pml or periodic in type hints.

PR #111, opened 2025-07-25
TYP: Rename and restructure the boundary objects and their states

kinimod0:
Issue #88

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "seaborn>=0.13.2",
     "matplotlib>=3.9.4",
     "rich>=13.9.4",
+    "typing-extensions>=4.14.1",
 ]
 
 [project.urls]
--- a/src/fdtdx/fdtd/initialization.py
+++ b/src/fdtdx/fdtd/initialization.py
@@ -318,7 +318,7 @@ def _init_arrays(
     if config.gradient_config is not None and config.gradient_config.recorder is not None:
         input_shape_dtypes = {}
         for boundary in objects.pml_objects:
-            cur_shape = boundary.boundary_interface_grid_shape()
+            cur_shape = boundary.interface_grid_shape()
             extended_shape = (3, *cur_shape)
             input_shape_dtypes[f"{boundary.name}_E"] = jax.ShapeDtypeStruct(shape=extended_shape, dtype=config.dtype)
             input_shape_dtypes[f"{boundary.name}_H"] = jax.ShapeDtypeStruct(shape=extended_shape, dtype=config.dtype)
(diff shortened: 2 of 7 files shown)


ymahlau:
This looks like a very well engineered solution for having a clean class structure with different boundaries. I actually did not know the typing.cast exists and learned something myself here. Well done and many thanks!

Some minor comments: 
- you can run the pre-commit pipeline locally with ```pre-commit run --all```. This will standardize the formatting, which is why the first check currently fails.
- fdtdx currently still supports python 3.11, which is why the import ```from typing import override``` does not work. Could you change this to the typing extension instead? You will also need to add the typing_extensions to the dependencies in pyproject.toml.

ymahlau:
@kinimod0 the way I usually think of state in the context of fdtdx is, that the state should contain everything that may change during a simulation. The full container holding this state is the ```fdtdx.ArrayContainer```, which contains the boundary states. For example, I think for the pml the boundary state only needs to contain the psi-coefficients. The other parameters bE, bH, cE, cH and kappa can be precomputed and saved as private attributes in the boundary object itself. You can do this similarly to how it is done in the [source object](https://github.com/ymahlau/fdtdx/blob/06fe946bffd1346c2decad16ed1f8131d9d4419f/src/fdtdx/objects/sources/source.py#L63) by overriding the place_on_grid method. 

I don't really think that there is a good way to separate the state and behavior of the dataclasses better than how it is currently done. But I am definitely open to suggestions, so if you have any ideas let me know. 

I also saw your #todo for the difference between init and reset state. In JAX, initializing an array and resetting is the exact same operation as long as the array lives on a single GPU. Since the boundary state is not sharded in the current implementation (which is not going to change in the near future), you can safely think about resetting and initializing as the same operation.

kinimod0:
Hi @ymahlau , I think there might be a misunderstanding. I'm not planning to put the state updater into the state class. My intention is to use composition, where the boundary class holds a `state_updater` instance that is separate from the state itself.

Thanks for pointing out the TODO and clarifying it 👍

I do have a question regarding the periodic boundary condition:
Why does `update_E_boundary_state` modify the H field?

kinimod0:
After testing out marker mixins and generics with typed parameter objects, I’ve concluded that they are probably overkill at this stage, so I’ll keep the current approach.

Does this cover everything you had in mind for this issue? If so, I’ll go ahead and mark the PR as ready for review.

ymahlau:
@kinimod0 Sorry for the late reply! I am currently travelling for a conference and will take a closer look at this when I find the time. At a first glance, the PR looks good though!

ymahlau:
@kinimod0 Thanks for the good work! I have no further comments and merged the PR.

Issue #103, opened 2025-07-21
Running simulation until condition is met


galcerte:
I have no idea if this is implemented, but I've been trying to find something akin to Meep's step functions ([https://meep.readthedocs.io/en/latest/Python_User_Interface/#run-and-step-functions](https://meep.readthedocs.io/en/latest/Python_User_Interface/#run-and-step-functions)) in FDTDX and I haven't been able to locate anything similar. I thought `custom_fdtd_forward` was the function that had anything like this but it doesn't seem to have it.

Basically, what I'm after is the possibility to stop the simulation if a user-specified condition is met, dependent on what the detectors read out. One example would be the time-averaged (over the last X wave half periods) Poynting flux changing less than some percentage, indicating a steady state rather than a transitory one.

galcerte:
Hmmm... I have a great interest in this feature, therefore I'm more than willing to work on it, but I'm not sure how you'd feel if a near first time contributor messed with the API. Are you accepting PRs on this, or should I wait?

galcerte:
The pull request I made to tackle this issue has been merged so I'm closing this issue (thought I couldn't do it). Thank you!

PR #117, opened 2025-08-05
Stopping simulation according to a user-specified condition

galcerte:
This PR attempts to solve issue #103. As suggested there, this here is a bit of a draft to obtain feedback on how this should be implemented. This is my first time when writing code using JAX that compiles to XLA, so said feedback is more than welcome.

On that note, I was wondering; Objects which are of class `DetectorState`, accessed by doing `arrays.detector_states["detector_name"]` are (so far) a dictionary with a key being the detector type name (`poynting_flux`, `phasor`, and so on) and the value being a `jax.Array`. As far as I understand, this array's shape changes with each time step, and I read that this may make it necessary to recompile functions/methods that might use it. I imagine recompiling this logic on every time step would be slow. But, is this true? Would I need to use arrays with a fixed shape to avoid recompilations? FDTDX already uses such arrays that vary their shape in time (detectors themselves) so I'm a bit confused on what to do in order to avoid recompilation and ensure good performance.

I also am throwing exceptions to indicate no compatibility with gradient computation, but I read that you can't do that inside JIT compiled code, and `custom_fdtd_forward` is supposedly compiled.

### DIFF
--- a/src/fdtdx/fdtd/stop_conditions.py
+++ b/src/fdtdx/fdtd/stop_conditions.py
@@ -1,0 +1,315 @@
+from abc import ABC, abstractmethod
+from typing import Self
+
+import jax
+import jax.numpy as jnp
+
+from fdtdx.config import SimulationConfig
+from fdtdx.core.jax.pytrees import TreeClass, autoinit, frozen_field, frozen_private_field
+from fdtdx.core.physics.metrics import compute_energy
+from fdtdx.core.wavelength import WaveCharacter
+from fdtdx.fdtd.container import DetectorState, ObjectContainer, SimulationState
+
+
+@autoinit
+class StoppingCondition(TreeClass, ABC):
+    """Abstract base class for defining custom stopping conditions in simulations.
+
+    This class provides the interface for implementing custom termination criteria
+    that can depend on simulation state, detector readings, field values, etc.
+    """
+
+    @abstractmethod
+    def setup(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> Self:
+        """Preparing condition for its use in the simulation."""
+        return self
+
+    @abstractmethod
+    def _validate(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> None:
+        """Pre-run validation; override in subclasses."""
+        pass
+
+    @abstractmethod
+    def __call__(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> jax.Array:
+        """Evaluate the stopping condition.
+
+        Args:
+            state (SimulationState): Current simulation state
+            config (SimulationConfig): Configuration of the simulation
+            objects (ObjectContainer): Objects in simulation
+
+        Returns:
+            jax.Array: Boolean scalar - True if simulation should continue, False if it should stop
+
+        Raises:
+            NotImplementedError: Must be implemented by subclasses.
+        """
+        raise NotImplementedError()
+
+
+@autoinit
+class TimeStepCondition(StoppingCondition):
+    """Default stopping condition based on maximum time steps.
+
+    This recreates the original behavior where simulation continues until
+    a specified end time is reached.
+    """
+
+    def setup(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> Self:
+        self._validate(state, config, objects)
+        return self
+
+    def _validate(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> None:
+        pass
+
+    def __call__(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> jax.Array:
+        """Check if simulation should continue based on time step.
+
+        Args:
+            state (SimulationState): Current simulation state.
+            config (SimulationConfig): Configuration of the simulation.
+            objects (ObjectContainer): Objects in simulation.
+
+        Returns:
+            jax.Array: Boolean scalar - True if current time < end_time, False otherwise
+        """
+        curr_time_step, _ = state
+        return config.time_steps_total > curr_time_step
+
+
+@autoinit
+class EnergyThresholdCondition(StoppingCondition):
+    """This condition stops a simulation when the total energy in the simulation volume
+    falls below a specified threshold. A minimum number of steps can be enforced
+    before convergence checks are applied, and a hard cutoff is imposed at
+    ``SimulationConfig.time_steps_total``. This condition is intended to be used with
+    pulsed sources, where the total amount of energy input in the volume is
+    finite, therefore total energy in the system is expected to converge towards zero eventually.
+
+    Attributes:
+        threshold (float, optional): Lower energy threshold for determining
+            simulation termination. Defaults to ``1e-6``.
+        min_steps (int, optional): The minimum number of time steps that must be
+            completed before convergence checking begins. Defaults to
+            ``0.1 * SimulationConfig.time_steps_total``.
+        max_steps (int, optional): The maximum number of time steps in a
+            simulation, regardless of the condition being fulfilled. Defaults to
+            ``SimulationConfig.time_steps_total``.
+    """
+
+    threshold: float = frozen_field(default=1e-6)
+    min_steps: int | None = frozen_field(default=None)
+    max_steps: int | None = frozen_field(default=None)
+
+    def setup(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> Self:
+        self = self.aset(
+            "max_steps", config.time_steps_total if self.max_steps is None else self.max_steps, create_new_ok=True
+        )
+        self = self.aset(
+            "min_steps",
+            int(round(config.time_steps_total * 0.1)) if self.min_steps is None else self.min_steps,
+            create_new_ok=True,
+        )
+        self._validate(state, config, objects)
+
+        return self
+
+    def _validate(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> None:
+        if self.threshold <= 0:
+            raise ValueError(f"Energy threshold must be positive, got {self.threshold}.")
+
+        if self.min_steps is not None and self.min_steps < 0:
+            raise ValueError(f"Minimum steps must be non-negative, got {self.min_steps}.")
+
+    def __call__(self, state: SimulationState, config: SimulationConfig, objects: ObjectContainer) -> jax.Array:
+        """Check if simulation should continue based on energy reaching a lower threshold.
+        Energy checks are only able to terminate the simulation after the time step is larger than ``min_steps``.
+
+        Args:
+            state (SimulationState): Current simulation state.
+            config (SimulationConfig): Configuration of the simulation.
+            objects (ObjectContainer): Objects in simulation.
+
+        Returns:
+            jax.Array: Boolean scalar - True if condition not met and curr_time_step < max_steps.
+        """
+        curr_time_step, arrays = state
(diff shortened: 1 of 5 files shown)


galcerte:
Okay, so I'm back working on this. Think I've identified a bit of a big problem with DetectorThresholdCondition aside from what you say; I was assuming that the detector reading array grew dynamically with every timestep, hence me always always the last element of this array, when in reality, that would just index the last ever reading, which wouldn't be filled until the last time step or last reading. I need to pass the current time step found in SimulationState to properly index the detector reading.

About the third issue you mentioned: in order to obtain the correct mapping, I imagine I need to pass the simulation's ObjectContainer when initializing the stopping condition object, right? Or is that overkill? That way I can access the detector object from there, get the switch (which if I understood correctly is stored as an attribute of the detector) and retrieve the attribute you mention.

ymahlau:
> Okay, so I'm back working on this. Think I've identified a bit of a big problem with DetectorThresholdCondition aside from what you say; I was assuming that the detector reading array grew dynamically with every timestep, hence me always always the last element of this array, when in reality, that would just index the last ever reading, which wouldn't be filled until the last time step or last reading. I need to pass the current time step found in SimulationState to properly index the detector reading.

Yes, this is correct :)

> About the third issue you mentioned: in order to obtain the correct mapping, I imagine I need to pass the simulation's ObjectContainer when initializing the stopping condition object, right? Or is that overkill? That way I can access the detector object from there, get the switch (which if I understood correctly is stored as an attribute of the detector) and retrieve the attribute you mention.

I think you almost have it the best way in the current code, just pass the name of the detector. Then you can get the correct DetectorState from the SimulationState in the __call__ function. 

Some further thoughts: 
- I think you don't need to pass the expected shape and end time step to the DetectorThresholdCondition. You can easily get this information from the array shapes. If you remove this, it makes the life of a user a little easier.
- What would be the main application of the detector stopping condition? The only application I can think of is stopping the simulation as soon as some oscillating steady state is reached at a detector. This, however, requires much more complicated logic than a simple "less than". Do you have a specific application in mind? If not this feature might not be used very often. You could even omit this feature for now from the PR and focus the effort on finalizing the other stopping conditions.

galcerte:
> The detector state values should be zero, not NaN.

Ah okay, I think I read something along those lines.

> I think the best way to realize this would be through the implementation of an detector, which integrates readings every time step. Then the stopping condition can access this detector reading. Please do not change the simulation state by adding more things to the tuple. This would result in more changes necessary for the custom autodiff in [reversible_fdtd](https://github.com/ymahlau/fdtdx/blob/b03a57e33a2db80fdc858eb9e59a8ccce55a2bcf/src/fdtdx/fdtd/fdtd.py#L16).

Okay, I see, makes total sense. I was reading what had to be changed if the carry value was something more than the SimulationState and it was a whole lot. I made FieldConvergenceCondition (now EnergyConvergenceCondition) require an EnergyDetector to read off of, so that should do it. I haven't made it check for `reduce_volume=True` but I will do that tomorrow. I think this StoppingCondition is now polished enough for you to take another look. I am not entirely sure if the code is completely JIT compilation friendly though...

> One simple way to also access the detector objects is by adding the object container as an additional argument to the call in stopping condition. Similar to the body_fn [here](https://github.com/ymahlau/fdtdx/blob/b03a57e33a2db80fdc858eb9e59a8ccce55a2bcf/src/fdtdx/fdtd/fdtd.py#L312), the cond_fn in the while then would need to receive a partial function where the objects argument is set statically.

In the end, I decided to do it as you said here. I accessed the Detector object through `objects`, the ObjectContainer instance used everywhere else already, fed to StoppingCondition through partial() in checkpointed_fdtd(). That way I get the mapping from time step to detector reading index that you mentioned. Mypy is giving me this error however: `Unexpected keyword argument "objects" for "__call__" of "StoppingCondition"`, even though I correctly specify an ObjectContainer as a parameter.

src/fdtdx/fdtd/stop_conditions.py
```diff
@@ -0,0 +1,160 @@
...
+    def __call__(self, state: SimulationState, objects: ObjectContainer) -> jax.Array:
+        """Evaluate the stopping condition.
+
+        Args:
+            state: Current simulation state (SimulationState)
+
+        Returns:
+            jax.Array: Boolean scalar - True if simulation should continue, False if it should stop
+        """
+        pass
```
ymahlau: better: raise NotimplementedError()

galcerte:
> Yes, I think this would still be very useful. For example, I believe Tidy3D also implements it this way.

Oh ok, that was very simple to do, I rewrote it like you said a couple of days ago.

> but I like the version with detectors more. In most applications a user would be interested to stop the simulation as soon as the detector measurement converged.

Yes, I agree, but previously I thought you had a different opinion on the matter, hence me asking.

> One way to test if a PoyntingFluxDetector/FieldDetector/EnergyDetector converged with a continous source would be to make a fourier transform from the time step (t-k*period) to (t-period) and then check if the last period matches the expected result. 

I've been thinking about this. Poynting flux detector readings (after the transitory phase) look more like the sum of a pure sine with double the frequency of the field and an arbitrary function of time. Since the Fourier transform is linear, that means you also get a sum of two terms; a term corresponding to the pure sine, and a term that is the transform of whatever is added to the pure sine. Therefore, doing something simple like just checking the frequency components with the highest magnitude would not be adequate, since they'd always be the same between different windows (the pure sine), at least with Poynting flux detectors. Doing something like checking the distance (doing an L2 norm for instance) between the spectra of the last period and the other k periods would better reflect convergence I believe, and is how I'm implementing this now, does this sound good to you?

Also, I only thought about doing this with a detector that registers every time step. What if the detector ingests an OnOffSwitch? So far my code depends on the period and the time step duration, which lets me get the number of samples per period. I think I can work with an OnOffSwitch that has `interval` set, but doing it with a switch that has arbitrary time steps selected, would be considerably harder...

src/fdtdx/fdtd/stop_conditions.py
galcerte: Done.

galcerte:
> Sounds good to me.

Great. Though I must add one thing: since the two subsets of the detector readings have different lengths ([t-k*period] -> [t-period], then [t-period, t]), one cannot do the L2 norm right away. I took the k previous periods and did the average of the detector reading over the k periods, so I get a signal that's the same length as the very last period. Since I do the average of windows of the reading with exactly the length of a period, the phase shouldn't be different between each window (which would distort the sinusoidal shape of the resulting average).

> I think you should first only think about a detector, which detects all time steps. The complicated parts we can still add in a later PR. First we should see if the general concept works.

Yeah that was my belief, I want to see this merged soon too... If that's the case, DetectorConvergenceCondition is essentially done, if you don't have any objection regarding its current impl. Only thing left are the tests, which I am currently working on.

ymahlau:
I took a closer look at the implementation and have some minor suggestions for improvements:

- Supplying the spp parameter to the Detector condition might be difficult for a user. It would be simpler to just let the user provide a WaveCharacter and compute spp internally. You can do this by adding a private frozen field for spp, which is initially None and then set the correct value in the place_on_grid or apply method. see [here](https://github.com/ymahlau/fdtdx/blob/6a828bad8820e0260f448314eddcc62ff8c22405/src/fdtdx/objects/detectors/detector.py#L108) for an example how to do this.
- For the detector convergence criterion, the minimum time step used should be max(min_time_steps, (k+1)*spp). Otherwise you will have incorrect readings because you are indexing outside of the array. 
- You probably noticed that a lot of the automatic test cases are failing in the PR, which you should address. I have not looked into detail why they fail, if you have specific questions just ask

galcerte:
> You can do this by adding a private frozen field for spp, which is initially None and then set the correct value in the place_on_grid or apply method. see [here](https://github.com/ymahlau/fdtdx/blob/6a828bad8820e0260f448314eddcc62ff8c22405/src/fdtdx/objects/detectors/detector.py#L108) for an example how to do this.

So I was thinking about this and I'm not quite sure what you meant; do you want the private field to go in DetectorConvergenceCondition, or to go in the Detector class? Because if I were to set the correct value in those methods, I'd have to put spp in Detector. Not sure if you meant that, or that I should do it in DetectorConvergenceCondition _like in those methods_.

galcerte:
> Sorry, this was a bit unclear from me: What I was thinking about was letting the user specify a WaveCharacter in the DetectorConvergenceCriterion (you can make this a public field like the spp is now). Then you can define spp in DetectorConvergenceCriterion as a private field: _spp: int = frozen_private_field(default=None)  # type: ignore.

> To actually set the correct _spp value, you can define a function similar to the place_on_grid function in the link i provided. In this method you compute the correct spp value and set the private attribute to this value. It is important that this function returns the self-value (the updated criterion class), because you cannot do in-place updates in JAX.

> I think the best way is to call this method instead of validate() in the actual fdtd function. Then the new function can internally also call validate. I am not sure about the naming, maybe something like "pre_simulation_setup". The call in the fdtd method might look like: new_criterion = criterion.pre_simulation_setup(arrays, objects). If you have better ideas for a name go with it.

> I also noticed that you code does not call validate if a default condition is provided. It would be better to call validate regardless of the provided condition. This makes future development easier.

This is all done in the last commits, and I went with setup() in the end.

By the way, what do you think of this?:

> Since we want to ease the user experience, I also thought of getting rid of end_steps, since that can be obtained from SimulationConfig. However, we are still left with min_steps which must be set by the user. Should I change min_steps such that it's the minimum time before checking the condition, and obtain min_steps internally as well? I imagine that would be more consistent.

Lastly, I fixed tests as much as I could, but the last 3 still fail. I'll try to fix those before I push again.

ymahlau:
> Since we want to ease the user experience, I also thought of getting rid of end_steps, since that can be obtained from SimulationConfig. However, we are still left with min_steps which must be set by the user. Should I change min_steps such that it's the minimum time before checking the condition, and obtain min_steps internally as well? I imagine that would be more consistent.

I think having an optional max_steps and min_steps parameter can be useful. But they should be optional, if no value is provided then max_steps should be determined by the config and min_steps internally from period and k.

galcerte:
Okay, I've finally made all the checks pass, tests should be good now. I've made an important change to DetectorConvergenceCondition; Instead of doing `min_steps = maximum(min_steps, (prev_periods + 1) * spp)`, I made it so the setup() method throws a ValueError if `min_steps < (prev_periods + 1) * spp`. This is done in order for the user to be more aware of what value should min_steps actually take, instead of silently setting it to a different value. Do you think this is fine? Or should I raise a Warning instead? Or just do `min_steps = maximum(min_steps, (prev_periods + 1) * spp)` as before?

src/fdtdx/fdtd/stop_conditions.py
```diff
@@ -0,0 +1,315 @@
...
+            config (SimulationConfig): Configuration of the simulation.
+            objects (ObjectContainer): Objects in simulation.
+
+        Returns:
+            jax.Array: Boolean scalar - True if condition not met and time < end_time.
+        """
+        curr_time_step, arrays = state
+        time_condition = curr_time_step < self.max_steps
+        min_steps_condition = curr_time_step < self.min_steps
+        total_energy = jnp.sum(arrays.E**2) + jnp.sum(arrays.H**2)
```
ymahlau: to avoid repeated code, please call compute_energy function here

ymahlau:
Apart from the minor comment above, this PR looks like it is ready to be merged! Please change the last minor things. Awesome work!

src/fdtdx/fdtd/stop_conditions.py
galcerte: Fixed.

PR #121, opened 2025-08-23
Added a new test

renaissancenerd:


ymahlau:
Hi @renaissancenerd, the test cases look good overall. I like how you used the mock module. A few minor points that need to be improved before merging the PR:
- Please run ```uv run pre-commit run --all``` locally to ensure that the formatting etc. is correct (see the failing check in the CI pipeline).
- Remove the empty file and commented, unused test cases
- For a clean github history in the main branch it would be good to avoid having many commits adding and then removing things again. You could reset your github history and add the new file in a single commit for example.
- Please use more descriptive git commit messages in the future. For example instead of "removing test" you could write "test: removed obsolete test_backward.py"

PR #132, opened 2025-09-06
Adding test_fdtd for fdtd module

renaissancenerd:


renaissancenerd:
@ymahlau Yes I see. I re-ran the tests and they do indeed take long. I will make the corresponding changes.

PR #136, opened 2025-09-09
adding sphinx docstrings for energy detector

renaissancenerd:


ymahlau:
Perfect, thanks!

ymahlau:
@renaissancenerd I actually just realized that there needs to be a column after the # for the comments. Could you please add this also?

src/fdtdx/objects/detectors/energy.py
```diff
@@ -12,24 +12,30 @@ class EnergyDetector(Detector):
...
-        reduce_volume (bool, optional): If True, reduces the volume data to a single energy value.
-            Defaults to False.
-        x_slice (float | None, optional): real-world positions for slice extraction. Defaults to None.
-        y_slice (float | None, optional): real-world positions for slice extraction. Defaults to None.
-        z_slice (float | None, optional): real-world positions for slice extraction. Defaults to None.
-        aggregate (str | None, optional): If "mean", aggregates slices by averaging instead of using position.
-            If None, mean is used. Defaults to None.
     """
 
+    # If True, returns energy measurements as 2D slices through the volume.
```
ymahlau, line 17: Should be ```#: If True, returns energy measurements as 2D slices through the volume.```

renaissancenerd: Sounds good.

PR #137, opened 2025-09-10
adding sphinx docstrings for conversion and core modules

renaissancenerd:


ymahlau:
Hi @renaissancenerd, I think there was a misunderstanding. Normal functions like ```export_stl``` should not be changed. Only subclasses of TreeClass and dataclasses should be changed, the rest should stay as is (especially functions). This applies to basically all of the functions in this PR.  From the changes in this PR, ```WaveCharacter``` and ```EnergyDetector``` are done correctly.

Another way to see if the docstring needs to be changed is to check if there is "Args" or "Attributes" written in the docstring. docstrings with Args should stay as is, docstrings with Attributes should be changed.

Please correct these points.

PR #142, opened 2025-09-12
adding test_misc for fdtd module

renaissancenerd:


ymahlau:
@renaissancenerd you have to add a function ```interface_slice``` to the Mock PML. The function is defined in the superclass of PML, the BaseBoundary: https://github.com/ymahlau/fdtdx/blob/669e30c048eb1c5aacf1525006efe1eab6794096/src/fdtdx/objects/boundaries/boundary.py#L107

PR #156, opened 2025-09-25
added argmax into unit, added test

sIHURs:
AC:
- Add an argmax function in unitful.py that correctly handles Unitful and other type data.
- Implement tests for argmax

### DIFF
--- a/src/fdtdx/units/unitful.py
+++ b/src/fdtdx/units/unitful.py
@@ -446,6 +446,9 @@ class Unitful(TreeClass):
     
     def prod(self, **kwargs) -> "Unitful":
         return prod(self, **kwargs)
+    
+    def argmax(self, **kwargs) -> np.number | np.ndarray | jax.Array:
+        return argmax(self, **kwargs)
         
 
 def align_scales(
@@ -1208,7 +1211,7 @@ def pow(x, y):  # type: ignore
 
 
 ## min #######################################
-def unary_fn(x: Unitful, op_str: str, *args, **kwargs) -> Unitful:
+def unary_fn(x: Unitful, op_str: str, *args, **kwargs) -> Unitful | ArrayLike:
     # TODO: handling of numpy arrays
     orig_fn = getattr(jax.numpy, f"_orig_{op_str}")
     new_val = orig_fn(x.val, *args, **kwargs)
@@ -1220,10 +1223,12 @@ def unary_fn(x: Unitful, op_str: str, *args, **kwargs) -> Unitful:
         if x_arr is not None:
             np_fn = getattr(np, f"{op_str}")
             new_static_arr = np_fn(x_arr, *args, **kwargs)
+    if not output_unitful_for_array(new_static_arr):
+        return new_val
     return Unitful(val=new_val, unit=x.unit, static_arr=new_static_arr)
 
 @overload
-def min(x: Unitful, *args, **kwargs) -> Unitful: return unary_fn(x, "min", *args, **kwargs)
+def min(x: Unitful, *args, **kwargs) -> Unitful: return unary_fn(x, "min", *args, **kwargs) # type: ignore
 
 @overload
 def min(x: PhysicalArrayLike, *args, **kwargs) -> jax.Array:
@@ -1240,7 +1245,7 @@ def min(x, *args, **kwargs):  # type: ignore
 
 ## max #######################################
 @overload
-def max(x: Unitful, *args, **kwargs) -> Unitful: return unary_fn(x, "max", *args, **kwargs)
+def max(x: Unitful, *args, **kwargs) -> Unitful: return unary_fn(x, "max", *args, **kwargs) # type: ignore
 
 @overload
 def max(x: PhysicalArrayLike, *args, **kwargs) -> jax.Array:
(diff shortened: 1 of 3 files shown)


src/fdtdx/units/unitful.py
```diff
@@ -1220,10 +1223,12 @@ def unary_fn(x: Unitful, op_str: str, *args, **kwargs) -> Unitful:
         if x_arr is not None:
             np_fn = getattr(np, f"{op_str}")
             new_static_arr = np_fn(x_arr, *args, **kwargs)
+    if not output_unitful_for_array(new_static_arr):
+        return new_val
     return Unitful(val=new_val, unit=x.unit, static_arr=new_static_arr)
 
 @overload
-def min(x: Unitful, *args, **kwargs) -> Unitful: return unary_fn(x, "min", *args, **kwargs)
+def min(x: Unitful, *args, **kwargs) -> Unitful: return unary_fn(x, "min", *args, **kwargs) # type: ignore
```
ymahlau: Instead of ignoring the typing error on the return type, it would be better to add a line like ```assert isinstance(result, Unitful)```.

src/fdtdx/units/unitful.py
```diff
@@ -1384,4 +1389,22 @@ def reshape(x: jax.Array, *args, **kwargs) -> jax.Array:
 @dispatch
 def reshape(x, *args, **kwargs):  # type: ignore
     del x, args, kwargs
-    raise NotImplementedError()
\ No newline at end of file
+    raise NotImplementedError()
+
+## argmax #######################################
+@overload
+def argmax(x: Unitful, *args, **kwargs) -> np.number | np.ndarray | jax.Array:
```
ymahlau: Unitful input should return a Unitful

src/fdtdx/units/unitful.py
```diff
@@ -1384,4 +1389,22 @@ def reshape(x: jax.Array, *args, **kwargs) -> jax.Array:
...
+    return unary_fn(x, "argmax", *args, **kwargs) # type: ignore
+
+@overload
+def argmax(x: PhysicalArrayLike, *args, **kwargs) -> np.number | np.ndarray | jax.Array:
```
ymahlau: Here you should make a more fine-grained differentiation: One function for jax Array input and output type, and another function for numpy input and output type.

src/fdtdx/units/unitful.py
ymahlau: In the test cases it is okay to put the type: ignore flag, but try to use this as little as possible in the source code itself. I used this myself a lot in the source code, which is not really good practice.

ymahlau: Only exception in the source code is the function annotated by ```@dispatch```, there is really no way around the type: ignore here.

src/fdtdx/units/unitful.py
```diff
@@ -1384,4 +1389,22 @@ def reshape(x: jax.Array, *args, **kwargs) -> jax.Array:
...
+    result_shape_dtype = jax.eval_shape(jnp._orig_mean, x, *args, **kwargs)  # type: ignore
+    if output_unitful_for_array(result_shape_dtype):
```
ymahlau: The behavior of outputting a unitful instead of an array is only necessary if a jax array is produced where none was present before. So for example jnp.asarray([1.0, 2.0]) converts python scalars to a jax array and needs this behavior. Here we already have a jax Array as input, so it is not necessary.

ymahlau:
You should add the following test cases:
- For Unitful input / jax Array input the argmax function should return the same output if called with jax.jit(argmax)(input)
- Unitful with numpy array value should have a numpy array as output
- Simple numpy array (no Unitful) as input
- If argmax is called on a python scalar, it should behave the same way as the standard numpy/jax argmax function. I am not sure if these functions would return index 0 as argmax or throw an exception because the input is not an array in this is case. Please find out what the original functions do and write a test case. 
- A bit more complicated scenario: if argmax is called within a jitted function on a Unitful with jax Array value and also static array, then the output should also have a static array. A function to test this might look like this:

```python
def test_static():
  def fn():
    x = Unitful(val=jnp.asarray([1.0, 2.0]), dim=...)
    if is_currently_jitting(): assert x.static_arr is not None
    result = argmax(x)
    if is_currently_jitting(): assert result.static_arr is not None
  jax.jit(fn)
```

sIHURs:
> You should add the following test cases:
> 
> * For Unitful input / jax Array input the argmax function should return the same output if called with jax.jit(argmax)(input)
> * Unitful with numpy array value should have a numpy array as output
> * Simple numpy array (no Unitful) as input
> * If argmax is called on a python scalar, it should behave the same way as the standard numpy/jax argmax function. I am not sure if these functions would return index 0 as argmax or throw an exception because the input is not an array in this is case. Please find out what the original functions do and write a test case.
> * A bit more complicated scenario: if argmax is called within a jitted function on a Unitful with jax Array value and also static array, then the output should also have a static array. A function to test this might look like this:
> 
> ```python
> def test_static():
>   def fn():
>     x = Unitful(val=jnp.asarray([1.0, 2.0]), dim=...)
>     if is_currently_jitting(): assert x.static_arr is not None
>     result = argmax(x)
>     if is_currently_jitting(): assert result.static_arr is not None
>   jax.jit(fn)
> ```
Regarding the last point: because in [code here](https://github.com/ymahlau/fdtdx/blob/1f8be8e76b4276d69794609da3c3bb6d9738afea/src/fdtdx/functional/jax.py#L50) the flag is being set directly, in unitful.py [code here](https://github.com/ymahlau/fdtdx/blob/1f8be8e76b4276d69794609da3c3bb6d9738afea/src/fdtdx/units/unitful.py#L1219) this ends up returning None, so we don’t get static_arr.

For this test, do the `assert x.static_arr is not None` and `assert result.static_arr is not None` checks have to be done during the JIT/compiling, or could we do them after calling jax.jit(fn), once it’s actually executed? Would that still satisfy the original test’s intent? Or should we better not change the flag in jax.py like do in another way?

PR #157, opened 2025-09-26
Fix method call to get frequency in PhasorDetector

brandmud:


ymahlau:
Thanks for catching this!

PR #162, opened 2025-10-15
added sign function

Shardy2907:


pyproject.toml
```diff
@@ -76,3 +76,8 @@ rocm = [
 [tool.pyright]
 include = ["src", "examples", "checks", "test"]
 exclude = ["**/.venv"]
+
+[dependency-groups]
+dev = [
+    "pytest>=8.4.2",
+]
```
ymahlau: There is already a development dependency group with pytest. Please use that. Do not change the pyproject.toml in this PR.

src/fdtdx/units/typing.py
```diff
@@ -75,8 +75,16 @@ class SI(Enum):
...
+if hasattr(np, "float128"):
+    PHYSICAL_DTYPES.append(np.float128)
+
+PHYSICAL_DTYPES = tuple(PHYSICAL_DTYPES)
+
+# Only add complex256 if available
+if hasattr(np, "complex256"):
+    PHYSICAL_DTYPES.append(np.complex256)
+
+PHYSICAL_DTYPES = tuple(PHYSICAL_DTYPES)
```
ymahlau: why is this necessary? is float128 and 256 not available on all systems?

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,71 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
 def argmin(x, *args, **kwargs):  # type: ignore
     del x, args, kwargs
     raise NotImplementedError()
+
+
+## sign #######################################
+from typing import overload as _typing_overload  # ensure name not shadowed
```
ymahlau: remove unused import

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,71 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+
+@overload
+def sign(x: np.ndarray) -> np.ndarray:
+    # NumPy sign then cast to numpy integer
+    y = np.sign(x)
+    return y.astype(np.int_)
+
+
+@overload
+def sign(x: int) -> int:
```
ymahlau, line 2078: pleases also add a typing hint for x: int | float.

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,71 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+                    y_s = np.int_(y_s)
+                new_static_arr = y_s
+        return Unitful(val=y_int, unit=EMPTY_UNIT, static_arr=new_static_arr)  # type: ignore
+
+
+@overload
+def sign(x: jax.Array) -> jax.Array:
+    # JAX original op via lax, then cast to integer dtype
+    y = jax.lax.sign(x)
+    return jax.lax.convert_element_type(y, jnp.int_)  # type: ignore
```
ymahlau: same as above, do not edit the output type

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,71 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+
+
+@overload
+def sign(x: np.ndarray) -> np.ndarray:
+    # NumPy sign then cast to numpy integer
+    y = np.sign(x)
+    return y.astype(np.int_)
```
ymahlau: same here

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,71 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+
+
+@dispatch
+def sign(x):  # type: ignore
+    del x
+    raise NotImplementedError()
+
+
+# --- make jnp.sign(...) use these overloads, preserving original ---
+if not hasattr(jnp, "_orig_sign"):
```
ymahlau: please take a look at how the current functions are patched: https://github.com/ymahlau/fdtdx/blob/unit/src/fdtdx/functional/patching.py

ymahlau: You do not need to add your own patching here, just do the same as for the other functions

src/fdtdx/units/typing.py
sIHURs: From my testing, these (np.float128, np.complex256) are not available on my MacBook Air M1.

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,63 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+@overload
+def sign(x: Unitful) -> Unitful:
+    # Drop units and return integer {-1,0,1}
+    if isinstance(x.val, np.ndarray | np.number):
+        # NumPy path
+        y = np.sign(x.val)
+        if isinstance(y, np.ndarray):
+            y_int: np.ndarray | np.number = y.astype(np.int_)
+        else:
+            y_int = np.int_(y)
```
ymahlau: you can delete the type conversion here, you can just use the output of np.sign

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,63 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+        return Unitful(val=y_int, unit=EMPTY_UNIT, static_arr=None)
+    else:
```
ymahlau: there should also be a case to handle the python int and float as value of a Unitful. Since this can be a lot of if-else with repeated code, you should rather just call sign(x.val), which calls the correct function from below.

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,63 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+        # JAX path (includes tracers)
+        y = jax.lax.sign(x.val)
+        y_int = jax.lax.convert_element_type(y, jnp.int_)  # type: ignore
```
ymahlau: again, no conversions necessary here

src/fdtdx/units/unitful.py
```diff
@@ -1608,3 +1608,63 @@ def argmin(x: np.ndarray, *args, **kwargs) -> np.ndarray:
...
+        if not isinstance(y_int, jax.Array):
+            raise Exception(f"This is an internal error: sign produced {type(y_int)}")
+        new_static_arr = None
+        if is_traced(y_int):
+            x_arr = get_static_operand(x)
+            if x_arr is not None:
+                y_s = np.sign(x_arr)
```
ymahlau: same goes for the static operand, just call sign(x_arr), which then gets redirected to one of the functions below

pyproject.toml
```diff
@@ -67,12 +67,15 @@ dev = [
     "myst_nb>=1.3.0",
 ]
 cuda12 = [
-    "jax[cuda12]>=0.4.36",
+    "jax[cuda12]>=0.7.0",
```
ymahlau, line 70: please do not make any changes to pyproject.toml in this PR

Shardy2907: Oh, yes my bad, will just add it to gitignore

Shardy2907: I had changed it in my earlier commit, so this time it was just setting it back to the version which is currently on github. so the 0.7.0 is the correct one. 0.4.36 was wrong. So I had just put it back to default.

PR #165, opened 2025-10-20
Modify and adjust Unitful comparison methods

sIHURs:
AC：
1. Output type matches input type (Unitful).
2. Comparison functions return correct results.

### DIFF
--- a/src/fdtdx/units/unitful.py
+++ b/src/fdtdx/units/unitful.py
@@ -32,6 +32,22 @@ like jnp.ones will return a Unitful instead of Array for scale optimization.
 """
 STATIC_OPTIM_STOP_FLAG = False
 
+"""
+Relative tolerance for floating point comparisons
+TODO: 
+    adjust RTOL_COMPARISON and ATOL_COMPARISON dynamically
+    based on the numeric precision (e.g., float16, float32, float64).
+    For now, these fixed values are sufficient for most use cases.
+
+| Precision Type | RTOL (Relative Tolerance) | ATOL (Absolute Tolerance) |
+| -------------- | ------------------------- | ------------------------- |
+| `float16`      | 1e-2 - 1e-3               | 1e-3 - 1e-4               |
+| `float32`      | 1e-4 - 1e-5               | 1e-6 - 1e-7               |
+| `float64`      | 1e-6 - 1e-8               | 1e-9 - 1e-12              | 
+"""
+RTOL_COMPARSION = 1e-5
+ATOL_COMPARSION = 1e-7
+
 
 @autoinit
 class Unit(TreeClass):
@@ -338,6 +354,15 @@ class Unitful(TreeClass):
         except Exception:
             return f"Unitful [{self.unit}]: {self.shape}"
 
+    def __bool__(self) -> bool:
+        if isinstance(self.val, (bool, np.bool_, jnp.bool_)):
+            return bool(self.val)
+        if isinstance(self.val, (np.ndarray, jnp.ndarray)):
+            if self.val.size != 1:
+                raise ValueError("Truth value of an array with multiple elements is ambiguous")
+            return bool(self.val.item())
+        return bool(self.val)
+
     def __repr__(self) -> str:
         return str(self)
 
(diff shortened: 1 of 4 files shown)


src/fdtdx/units/unitful.py
```diff
@@ -1054,25 +1065,34 @@ def le(x, y):  # type: ignore
...
+) -> Unitful:
     x_align, y_align = align_scales(x, y)
-    new_val = x_align.val == y_align.val
+
+    if isinstance(x_align.val, jnp.ndarray) or "jax" in str(type(x)):
+        new_val = jnp.isclose(x_align.val, y_align.val, rtol=1e-5, atol=1e-8)
+    elif isinstance(x_align.val, np.ndarray | np.number):
+        new_val = np.isclose(x_align.val, y_align.val, rtol=1e-5, atol=1e-8)
+    elif isinstance(x_align.val, float | int):
+        new_val = abs(x_align.val - y_align.val) <= 1e-8 + 1e-5 * abs(y_align.val)
```
ymahlau: Here, we should not alter the behavior of numpy / jax / python. Please use the numpy.equal / jnp.equal methods here. For python just use ==. 

Additionally please change the following: If the value of one input is a jax array and the other is a numpy array, the result should be a jax array. Basically the hierarchy should be jax > numpy > python. If there is a combination of types, the result should be the greater type in the hierarchy. You probably don't need any manual conversions, jnp.equal should have this behavior by default with python / numpy inputs for example.

src/fdtdx/units/unitful.py
```diff
@@ -1054,25 +1065,34 @@ def le(x, y):  # type: ignore
...
+    else:
+        raise TypeError(f"Cannot compare values of type {type(x)} and {type(y)}: x={x}, y={y}")
+
     new_static_arr = None
     if is_traced(new_val):
         x_arr = get_static_operand(x_align)
         y_arr = get_static_operand(y_align)
         if x_arr is not None and y_arr is not None:
-            new_static_arr = x_arr == y_arr
+            new_static_arr = np.isclose(x_arr, y_arr, rtol=1e-5, atol=1e-8)
```
ymahlau: Same goes here

src/fdtdx/units/unitful.py
ymahlau: Since we are ovveriding these methods, you can also rename this function from "eq" to "equal" to reflect the numpy / jax functions. Same goes for the other functions.

src/fdtdx/units/unitful.py
```diff
@@ -1054,25 +1065,34 @@ def le(x, y):  # type: ignore
 def eq(
     x: Unitful,
     y: Unitful,
-) -> np.bool | np.ndarray | jax.Array | bool:
+) -> Unitful:
     x_align, y_align = align_scales(x, y)
-    new_val = x_align.val == y_align.val
+
+    if isinstance(x_align.val, jnp.ndarray) or "jax" in str(type(x)):
```
ymahlau: I think the best way to check if the input is a jax array should be "isinstance(x_align.val, jax.Array)"

ymahlau:
Could you also add a \_\_bool__  magic method to unitful? This would allow using a unitful in an if-clause like this

```python
x = Unitful(val=True, ...)
if x:
   # do something
```

ymahlau:
Thanks for the good work! This already looks really good, just the few things I mentioned above need changing

src/fdtdx/units/unitful.py
sIHURs: this change because after running the scale alignment, the values end up slightly off.

I ran into this situation while writing the test.

```python
def test_le_untiful_jitted():
    """Test less than or equal within a JIT-compiled function when input as unitful"""

    length_unit = Unit(scale=-2, dim={SI.m: 1})
    lengths1 = Unitful(val=jnp.array([12.0, 25.5, 18.3, 30.1, 22.7]), unit=length_unit)
    lengths2 = Unitful(val=jnp.array([12.0, 14.5, 100.3, 302.1, 2.7]), unit=length_unit)
    print("uq unit scale: ", lengths1.unit.scale)
    print("length1", lengths1.val)
    print("length2", lengths2.val

    result = lengths1 <= lengths2
 
    assert isinstance(result, Unitful)
    assert isinstance(result.val, jax.Array)
    assert jnp.allclose(result.val, jnp.array([True, False, True, True, False]))
```
After optimizing and aligning the scale, the two input values became:
```python
x_align: [1.2       2.55      1.8299999 3.01      2.2700002]
y_align: [1.1999999  1.4499999 10.030001  30.21       0.27]
```
The first element in Unitful y lost some precision, so using Python’s == would return False directly. That’s why I used something like this in le(), ge(), and eq().
```python
new_val = xx.logical_or(xx.isclose(...), x < y)
```

Or we could increase the precision of the variables where needed, using float64, but that would use more memory.

ymahlau: Ah I see, this makes sense, thanks for the explanation. Then you can leave it with the floating point comparison and just change the other points I mentioned. 

Since we then leave the floating point values in the Code, please do not write constants like 1e-8 directly in the functions. Rather use a constant RTOL_COMPARISON = 1e-8 at the beginning of the file and then use this throughout the comparison functions. Also please experiment a little bit to make the tolerance as small as possible (if the comparison fails in rare occations then its fine, it should just mostly give the same result).

Please also add a todo that in the future we should adjust the tolerances based on the datatype (float16, float32, float64), but for now don't do this. This is not necessary for now.

src/fdtdx/units/unitful.py
```diff
@@ -1054,25 +1098,34 @@ def le(x, y):  # type: ignore
...
     x: Unitful,
     y: Unitful,
-) -> np.bool | np.ndarray | jax.Array | bool:
+) -> Unitful:
     x_align, y_align = align_scales(x, y)
-    new_val = x_align.val == y_align.val
+
+    if isinstance(x_align.val, jax.Array):
+        new_val = jnp.isclose(x_align.val, y_align.val, rtol=RTOL_COMPARSION, atol=ATOL_COMPARSION)
+    elif isinstance(x_align.val, np.ndarray | np.number):
```
ymahlau: What would happen here, if y.val is a jax Array and x.val is a numpy array? In that case the result should be a jax Array, but in your implementation it would be a numpy array, correct? 

You should make a check if either x or y value is a jax array. If so use jnp.isclose. Then check if either of them is a numpy array, if so use np.isclose. In all other cases, use the python comparison.

sIHURs: Thanks for the reminder, I’ve updated it in the new commit.

Issue #164, opened 2025-10-19
Adding alpha as a parameter to BoundaryConfig and alpha being scaled with PML depth


galcerte:
I have noticed that $\alpha$, the complex frequency shifting term, cannot be set, even though it is exposed in `PerfectlyMatchedLayer` as `alpha`, unlike the thickness of the PML and the $\kappa$ at the start and end of the PML region. Is this because of anything in particular, or can such a parameter be added without much trouble? I have experienced signifcant gains in accuracy by varying this parameter and haven't noticed anything weird so far, so I think it shouldn't be hard to add this.

While reading on how to reduce reflections in PMLs, I stumbled across a paper [1] that mentioned: "To reduce the discretization error, the parameters $\sigma_i$, $\kappa_i$ are scaled such that they are 0 and 1 at the PML/working volume interface, respectively, and are maximum at the exterior boundary [1, 2]. However, $\alpha_i$ is not scaled in the same manner. Specifically, to reduce reflection error of evanescent modes $\alpha_i$ must be non-zero at the front boundary interface. However, for the CFS-PML to absorb purely propagating modes at low frequency, $\alpha_i$ should actually decrease to zero away from the boundary interface [10]." So looking into this, I noticed that in FDTDX, `alpha` is constant with PML depth, unlike `kappa`. Varying $\alpha$ with PML depth might yield more absorption in some situations, so it's probably worth adding this feature.

I am willing to open a PR to tackle this issue.

[1]: Roden, J. A. and Gedney, S. D., “Convolution PML (CPML): An efficient FDTD implementation of the CFS–PML for arbitrary media,” Microwave and Optical Technology Letters 27(5), 334–339 (2000).

ymahlau:
I agree that this should not be difficult to implement. Go for it if you want! I think we should keep the current static alpha as a default value, because it seems to work well for most applications.

PR #166, opened 2025-10-23
Linear grading of alpha and exposing respective knobs through BoundaryConfig

galcerte:
This pull request tackles issue #164. It implements:
- A linear variation of alpha with PML depth, like kappa, instead of having it be constant throughout the PML.
- Replaced `alpha` in `PerfectlyMatchedLayer` for `alpha_start` and `alpha_end`, again, similarly to kappa.
- Exposed `alpha_start` and `alpha_end` through `BoundaryConfig` to let the end user tune alpha as they wish.

Default values have been kept, through letting both `alpha_start` and `alpha_end` be equal to `1e-8`.


ymahlau:
Awesome! thanks for the contribution

Issue #167, opened 2025-10-26
Refactor Object Placement Code


ymahlau:
In line with the ongoing effort to develop a GUI for fdtdx, we need to refactor the object placement code. Currently this code is a bit of a mess, not tested and lets some features to be desired. 

The following changes need to be done:
- The coordinate system should have its origin at the center of the simulation volume, not the lower front left corner. This requires some changes to the grid slice tuple of an object (see [here](https://github.com/ymahlau/fdtdx/blob/9ce79b760c497723c3efb2772e1ca68a65e32750/src/fdtdx/objects/object.py#L234)). Most likely it only needs to be changed when set in place_objects (see points below).
- SimulationObjects should have the option to set a position directly in the constructor. Similar to [partial_real_shape](https://github.com/ymahlau/fdtdx/blob/9ce79b760c497723c3efb2772e1ca68a65e32750/src/fdtdx/objects/object.py#L209), add an attribute called partial_real_position. This is only necessary for real position, we do not need a grid position. Since this makes [RealCoordinateConstraint](https://github.com/ymahlau/fdtdx/blob/9ce79b760c497723c3efb2772e1ca68a65e32750/src/fdtdx/objects/object.py#L175) obsolete, you can delete it. 
- The constraints defined in object.py (see [here](https://github.com/ymahlau/fdtdx/blob/9ce79b760c497723c3efb2772e1ca68a65e32750/src/fdtdx/objects/object.py#L67)) should not save a whole SimulationObject as object and other_object. Instead the attribute should be a string defining the name of the respective object.
- Because of the aforementioned change, the [place_object](https://github.com/ymahlau/fdtdx/blob/9ce79b760c497723c3efb2772e1ca68a65e32750/src/fdtdx/fdtd/initialization.py#L29) method needs to take in a list of objects as an additional argument.
- the very messy [_resolve_object_constraint](https://github.com/ymahlau/fdtdx/blob/9ce79b760c497723c3efb2772e1ca68a65e32750/src/fdtdx/fdtd/initialization.py#L372) function should be cleaned up:
  - it should also take in a list of objects as an argument
  - Output should be a dict[str, SliceTuple3D | None], where None marks that the position / size of an object was not defined properly.
[...]

PR #169, opened 2025-10-29
refactoring object placement code #167

renaissancenerd:


src/fdtdx/fdtd/initialization.py
```diff
@@ -27,97 +26,106 @@
...
-    # place objects on computed grid positions
+    # Step 2: Aggregate errors and raise if needed
+    failed = {name: msg for name, msg in errors.items() if msg}
+    if failed:
+        formatted = "\n".join(f"  - {name}: {msg}" for name, msg in failed.items())
+        raise ValueError(f"Failed to resolve object constraints:\n{formatted}")
+
+    # Step 3: Convert name → object for placement
+    object_map = {obj.name: obj for obj in objects}
+    volume_obj = next((o for o in objects if o.name.lower() == "volume"), None)
```
ymahlau: This should rather test for isinstance(o, SimulationVolume). Also should throw an error if multiple simulation volumes are given.

src/fdtdx/fdtd/initialization.py
```diff
@@ -27,97 +26,106 @@
...
+        ("PositionConstraint" | "SizeConstraint" | "SizeExtensionConstraint" | "GridCoordinateConstraint")
     ],
-    key: jax.Array,
+    key: "jax.Array",
 ) -> tuple[
-    ObjectContainer,
-    ArrayContainer,
-    ParameterContainer,
-    SimulationConfig,
+    "ObjectContainer",
```
ymahlau: Why do we need the apostrophs here?

src/fdtdx/fdtd/initialization.py
```diff
@@ -370,327 +378,250 @@ def _init_params(
...
+        if all_done:
+            break
+
+        # Stop early if no changes in several iterations (stuck)
+        if not changed:
+            stable_iterations += 1
+            if stable_iterations > 10:
+                for o in object_map.values():
+                    if errors[o.name] is None:
+                        errors[o.name] = "Unresolved: constraints could not converge."
```
ymahlau: Please make the error messages a bit more descriptive. In the best case they should give the user clear instructions how to fix the errors (If the exact error is known).

src/fdtdx/fdtd/initialization.py
```diff
@@ -370,327 +378,250 @@ def _init_params(
...
+def _finalize_resolution(objects, shape_dict, slice_dict, errors):
+    """Convert lists into SliceTuple3D and fill None for unresolved entries."""
+    resolved: dict[str, "SliceTuple3D" | None] = {}
+    for o in objects:
+        if errors[o.name] is not None:
+            resolved[o.name] = None
+            continue
+        if any(v is None for v in shape_dict[o]) or any(any(v is None for v in pair) for pair in slice_dict[o]):
+            resolved[o.name] = None
+            errors[o.name] = "Incomplete constraint resolution"
```
ymahlau, line 711: Same here

src/fdtdx/fdtd/initialization.py
```diff
@@ -49,75 +40,97 @@ def place_objects(
...
             - ParameterContainer with device parameters
             - Updated SimulationConfig
             - Dictionary with additional initialization info
+
+    Raises:
+        ValueError: If constraint resolution fails for one or more objects.
     """
-    slice_tuple_dict = _resolve_object_constraints(
-        volume=volume,
+    # Step 1: Resolve constraints into grid slices
```
ymahlau, line 75: Can you add a sanity check here that tests if all objects have unique names? If there are two or more objects with the same name it should return an Exception to the user.

src/fdtdx/fdtd/initialization.py
```diff
@@ -370,327 +383,281 @@ def _init_params(
 
 
 def _resolve_object_constraints(
```
ymahlau: Can you please rename this function without an underscore at the beginning? This is the function that would be called by the GUI and needs to be accessible by a user. The underscore implies it would be a private function

src/fdtdx/fdtd/initialization.py
```diff
@@ -370,327 +383,281 @@ def _init_params(
...
+
+    Args:
+        object_map: Dict mapping object names to SimulationObject instances.
+        constraints: Sequence of constraint instances referencing object names.
+        shape_dict: Dict tracking (x, y, z) shape values in grid units.
+        slice_dict: Dict tracking grid slice start/end positions for each axis.
+        errors: Dict collecting user-friendly error messages.
+    """
+    MAX_ITER = 1000
+    stable_iterations = 0
```
ymahlau: You do not need a stable iterations counter. If nothing changed for a single iteration, then it will not change anymore. So you can already directly stop using the "if not changed" logic below directly.

src/fdtdx/fdtd/initialization.py
```diff
@@ -370,327 +383,281 @@ def _init_params(
...
+
+                elif isinstance(c, SizeConstraint):
+                    changed |= _apply_size_constraint(c, obj, other, shape_dict)
+
+                elif isinstance(c, SizeExtensionConstraint):
+                    changed |= _apply_size_extension_constraint(c, obj, other, shape_dict)
+
+                elif isinstance(c, GridCoordinateConstraint):
+                    changed |= _apply_grid_coordinate_constraint(c, obj, shape_dict, slice_dict)
+
```
ymahlau, line 552: Please add an else case without condition here. This should test if we by accident have a "c" here that does not fit the previous if-cases. If so throw an error that c is unknown.

src/fdtdx/fdtd/initialization.py
```diff
@@ -370,327 +383,281 @@ def _init_params(
...
+    Iteratively apply all constraints until shapes and positions converge.
+
+    Args:
+        object_map: Dict mapping object names to SimulationObject instances.
+        constraints: Sequence of constraint instances referencing object names.
+        shape_dict: Dict tracking (x, y, z) shape values in grid units.
+        slice_dict: Dict tracking grid slice start/end positions for each axis.
+        errors: Dict collecting user-friendly error messages.
+    """
+    MAX_ITER = 1000
```
ymahlau: Please make max_iter an argument to the function itself. The default argument should be DEFAULT_MAX_ITER, which can be defined at the top of the file.

ymahlau:
I did not look into detail into _apply_position_constraint, _apply_size_constraint etc. For testing these it would be best to write unit tests with some objects and call these functions directly in the test.

ymahlau:
Additionally, you will need to adjust the existing test cases, because the internals changed. try to stay away from using mocks as much as possible and preferably define actual objects in the test cases

Issue #172, opened 2025-10-31
Excessive PML reflection


galcerte:
While investigating how to configure the PML to reduce reflections (my target is -60 to -80 dB, and I got about -35 dB, more info here [https://github.com/ymahlau/fdtdx/discussions/161](https://github.com/ymahlau/fdtdx/discussions/161)) I looked into the implementation itself and I found a few worrying things. Starting off with the light one:

- In `perfectly_matched_layer.py`, when computing `cE` and `cH`, the first factor of the expression is `bE - 1` and `bH - 1`, functions of the form `jnp.exp(x) - 1`, and for small values of `x`, it is preferable to use `jnp.expm1(x)` since it has much higher precision (1). In this case, the argument `x` is `-self._config.courant_number * (sigma_E / kappa + alpha)`, which is small at the interface between PML and vacuum at the default value `alpha_start=1e-8`; for float32 the error is about a factor of 12 (=1.2e-7/1e-8), so pretty significant.
- In the methods `update_E` and `update_H`, which update the fields in the PML regions, one can see that the fields that these methods take as input are divided by `kappa`. However, if we check where these methods are called, we end up in `update.py`, where we can see that, before the fields are passed to these methods, they are already updated everywhere in the volume. Contrasting how these fields are updated versus what Schneider, chapter 3.12 (2) says, we can observe several mismatches (we shall focus on the lossless electric field `E` but the magnetic field is very similar);
    -  The next time step's field is divided by `kappa` as a whole. However, in Schneider, equation (11.110), one can see that `kappa` should not divide everything before summing `phi_E`. It shouldn't divide the current time step's electric field (observe the term `factor * E` in line 73 of `update.py`).
[...]

Issue #84, opened 2025-06-27
Pml performance issue


yff-me:
Hi! I'm testing this solver on A100 and found the performance highly dependent on the boundaries:
- With all `periodic` boundaries, it can run up to 6000 Million Cells per second.
- With all `pml` boundaries, the performance goes down to 800 Million Cells per second.

To my knowledge, this is really different from traditional FDTD simulation software. Any chance to improve the performance?

ymahlau:
Hi, thanks for bringing this up! 

The way it is currently implemented is that after each E / H update, the pml updates are run. Currently this is quite inefficient, because all six pml updates are run sequentially. 

One simple performance improvement would be to run the two opposite pml in parallel and therefore only have three sequential pml updates instead of six. It is currently not possible to run all six pml in parallel, because the pml overlap. 

The most efficient way to do this would be to split the E / H field into different arrays for the inside and pml regions. Then one could also implement special pml variants for the corner and run all pml in parallel. However, this would need considerable programming effort.

There might also be more efficient pml variants compared to the current convolutional PML implementations. 

I will implement the simple improvement in the near future.

PR #175, opened 2025-11-02
PML implementation overhaul

galcerte:
This PR tackles issue #172. The initial commit on here only fixes the first two of the three points mentioned there. The last point on that issue is what will probably take the most time to complete, that is, updating the fields once across all PMLs.

Even though now the PML object updates the fields according to literature, in my testing, the PML is still behaving rather weirdly, even if no PML volumes in different axis overlap. This is the script I am using to test this out: [https://gist.github.com/galcerte/d062bc8446a766be5e03a2d8f4f5fecc](https://gist.github.com/galcerte/d062bc8446a766be5e03a2d8f4f5fecc). Maybe it's worth including this script in the `checks/` folder?

Other things that still need to be done is to check if the PMLs work fine with lossy materials inside them.


ymahlau:
I will take a look at this later. I think for implementing non-overlapping PML it would be best to generalize the PML implementation to multiple directions (x, y, z, either positive / negative), which can be specified by the user. This would allow specifying a separate PML object for the corner and sides. Then the get_boundary_from_config can automatically generate the appropriate objects (same pml as before on the side, separate pml object for corners). Specifying the position of these objects by hand would probably be tedious, but should be easy to automate.

ymahlau:
As discussed in #84, the updates in the pml are not carried out in parallel, because the pmls return the complete array, which makes it difficult for the jit optimizer to identify that the updates can actually be performed in parallel. There are ways to do this better (constructing a new array from the updated pmls and the non-updated regions through concatenation), but this may be a lot of work to implement.  Could you add the new implementation and also keep the old one such that a user can decide if they want the higher accuracy or faster runtime?

galcerte:
I was thinking of another way to go about it; instead of updating 26 times, what about taking all the grid slices, taking in all the alpha, kappa, and sigmas (as well as the bE and bH, cB and cH coefficients; basically everything in PMLBoundaryState), concatenating _those_ and updating _just once_? The problem with this is that the updating would probably have to be done elsewhere, as one single `update_E` or `update_H` method in each PML object does not have the full picture by accessing their own object's attributes; probably a different helper function distinct from all those methods would do the trick. Then each PML object would be in charge of just storing the previous state in PMLBoundaryState, passing them to the function, and letting it concatenate all those arrays and update the fields using that.

Or maybe it would be worth making just 1 PML object for all PML boundaries. That way the `update_E` and `update_H` interface would be intact, and its usage in `update.py`. But that would need more work in `initialization.py`.

> Could you add the new implementation and also keep the old one such that a user can decide if they want the higher accuracy or faster runtime?

I can do that, but honestly I also think it's worth putting in the effort to make this implementation be both accurate and fast.

ymahlau:
Yes, if you can manage to make the implementation both fast and accurate then we do not need the old parts anymore. However, I think it would be difficult to make this work in all cases, especially when using multiple GPU. For these cases we should keep the old implementation as an option.

The options you are suggesting all eventually come down to the same thing I was saying above:
- you will need to give the pml objects the individual slices of the E/H field and retrieve the updated E/H slices. jax will do that in parallel automatically.
- create a 3d array of the original simulation volume size from these arrays using jnp.concatenate and use the old E/H field where no updates happened (this should be one big slice in the middle of the simulation volume. If there is a combination of periodic and pml boundaries this will need to be adjusted.

Issue #174, opened 2025-11-02
Improvements to existing methods in fdtdx.units.unitful


ymahlau:
The existing methods in units.unitful, which I implemented in the past, need some improvements:

- For add/multiply/subtract/division: 
  - add proper methods for numpy in all combinations with python / numpy types 
  - same for any combination of jax array as x and numpy/python as y-value (and the other way around)
  - combination of python values (e.g x is complex and y is int etc.)
- matmul needs numpy implementation
- pow should have more fine-grained typing (e.g. float, int -> float)
- behavior of min/max/mean/sum/prod should be changed: 
  - if input is jax.Array, output should be jax.Array. For all other input types use the numpy functions, not the jax functions.
  - for Unitful input: if the unitful value is a python value, then unary_fn should call the numpy functions, not the jax functions
- abs_impl needs numpy implementation
- astype/squeeze/reshape/: same points as min/max/... except that for these function python values are not valid inputs

@sIHURs could you implement these improvements?

PR #183, opened 2025-11-14
Unit Functions Improvements requested in issue #174

sIHURs:
This PR tackles issue https://github.com/ymahlau/fdtdx/issues/174.

### DIFF
--- a/tests/units/test_unitful.py
+++ b/tests/units/test_unitful.py
@@ -451,16 +451,17 @@ def test_le_untiful_jitted():
 
 
 def test_le_jitted_static():
+    """Test less than or equal within a JIT-compiled function when input as unitful with static arrays"""
     length_unit = Unit(scale=-2, dim={SI.m: 1})
     lengths1 = Unitful(
-        val=jnp.array([12.0, 5.5, 18.3, 30.1, 22.7]),
+        val=jnp.array([12.0, 5.5, 18.3, 30.1]),
         unit=length_unit,
-        static_arr=np.array([12.0, 5.5, 18.3, 30.1, 22.7]),
+        static_arr=np.array([12.0, 5.5, 18.3, 30.1]),
     )
     lengths2 = Unitful(
-        val=jnp.array([12.0, 14.5, 1.3, 3012.1, 22.7]),
+        val=jnp.array([12.0, 14.5, 1.3, 3012.1]),
         unit=length_unit,
-        static_arr=np.array([12.0, 14.5, 100.3, 3012.1, 22.7]),
+        static_arr=np.array([12.0, 14.5, 100.3, 3012.1]),
     )
 
     def fn(x: Unitful, y: Unitful) -> Unitful:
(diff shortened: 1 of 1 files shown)


src/fdtdx/units/unitful.py
```diff
@@ -1380,17 +1617,83 @@ def pow(x: Unitful, y: int) -> Unitful:
...
+
+
+@overload
+def pow(x: float | int, y: complex) -> complex:
+    return x**y
+
+
+@overload
+def pow(x: complex, y: complex) -> complex:
+    return x**y
```
ymahlau, line 1627: you can merge the two functions above using (x: complex | float | int, y: complex)

src/fdtdx/units/unitful.py
```diff
@@ -1399,20 +1702,21 @@ def pow(x, y):  # type: ignore
...
-        if not isinstance(new_val, np.ndarray) and not isinstance(new_val, np.number):
+    # handling of jax Array
+    if isinstance(x.val, jax.Array):
+        orig_fn = getattr(jax.numpy, f"_orig_{op_str}")
+        new_val = orig_fn(x.val, *args, **kwargs)
+        if not isinstance(new_val, jax.Array):
             raise Exception(f"This is an internal error: {op_str} produced {type(new_val)}")
         if new_val.dtype not in PHYSICAL_DTYPES:
             return Unitful(val=new_val, unit=EMPTY_UNIT, static_arr=None)
         return Unitful(val=new_val, unit=x.unit, static_arr=None)
```
ymahlau: This return statement here is problematic (I think this might have slipped my review in an earlier PR?). It will always prevent saving the static array if the input is a jax Array. Instead, the computation below should be performed and saved as static array. 

I think the easiest way to realize this is by adding an else-statement and putting line 1716 and 1717 in there.

src/fdtdx/units/unitful.py
```diff
@@ -1430,15 +1734,35 @@ def min(x: Unitful, *args, **kwargs) -> Unitful:
...
 
 
+"""
+PhysicalArrayLike = Union[
+    float,
+    complex,
+    jax.Array,
+    np.number,
+    np.ndarray,
+"""
```
ymahlau: please remove temporary comments

src/fdtdx/units/unitful.py
ymahlau: to clarify, i mean the return statement in line 1410 / 1713 is problematic: ```return Unitful(val=new_val, unit=x.unit, static_arr=None)```

sIHURs: In this PR I reordered the handling of jax.Array and NumPy values, but I forgot to change the new_static_arr part accordingly. Thanks for pointing it out — the function has been fixed now.

ymahlau:
Hi @sIHURs, I just realized that there are two test cases failing in the test_unitful.py. This might have already happenend in previous PRs, but could you please investigate?

Issue #108, opened 2025-07-24
Lossy Material: Simulation and analytical solution disagree quite a lot


ymahlau:
In the [example](https://ymahlau.github.io/fdtdx/notebooks/simulate_lossy_material/) for lossy materials, the analytical solution and simulation disagree by quite a large margin. Investigation why this is happening is necessary.

PR #185, opened 2025-11-14
Refactor code to fix PML boundary issue

rachsmith1:
Hello @ymahlau , I have been following along closely with the issues related to the implementation of the PML (see #108 #161 #172 and the PR #175 opened by @galcerte). I agree that the current implementation of PML is physically incorrect and causes excessive reflections. 

The issue overall seems to be as follows: how to implement a physically correct PML in a way that is both memory efficient and does not excessively slow down the simulation time. This PR offers a potential solution. I don't pretend it is the best possible solution so please let me know what you think.

This solution works as follows: instead of the boundary objects having their own fields that are updated separately and in sequence (leading to fields that are updated multiple times if the boundary objects overlap), instead the alpha, kappa, and sigma PML parameters, as well as the auxiliary fields, are available as global arrays (similar to the E and H fields, or the permittivity). These arrays are then _modified_ by the boundary objects in [initialization.py](https://github.com/XanaduAI/fdtdx/blob/boundary-layer-refactor/src/fdtdx/fdtd/initialization.py#L350). Then, in [update.py](https://github.com/XanaduAI/fdtdx/blob/boundary-layer-refactor/src/fdtdx/fdtd/update.py), all of the fields everywhere are updated **once** and the areas with modified PML parameters behave in a physically correct way. Since there is only one update per time step for the whole simulation, this is quite fast.

Pros:
- Physically correct and fast.
- There should be no difference on the user end as far as choosing boundaries of different types and on different axes of the simulation. They would have more options for tuning the alpha, kappa, and sigma parameters of the PML. 
- I also thought of adding functionality for the boundaries to be able to modify the electric and magnetic conductivity arrays. This would allow implementation of an adiabatic absorber type boundary such as [the one in Tidy3D](https://docs.flexcompute.com/projects/tidy3d/en/latest/api/_autosummary/tidy3d.Absorber.html).

Cons:
- Because the alpha, kappa, sigma, and auxiliary field arrays cover the entire simulation, well on the one hand it means the user can specify boundary conditions anywhere. But on the other hand for areas without boundaries (i.e. most of the simulation) you just have large arrays with mostly useless values (all zeros, except for kappa which by default is all ones). This seems highly memory inefficient. I was thinking some implementation of sparse arrays might help.

Questions:
- I have only tested using forward fdtd sims. I don't believe this should break anything with backward sims but have not tested this yet.
- I don't know how well this works with multiple GPUs.

As an example of how this PML implementation affects simulation behavior, I have also added my lossy waveguide width sweep analysis as an additional [example](https://github.com/XanaduAI/fdtdx/blob/boundary-layer-refactor/examples/width_sweep_analysis.py). In this analysis, I setup a straight waveguide with sidewall absorbers that have electric conductivity. There is a mode source and multiple mode monitors. I compare the analytic loss as calculated by the complex effective index that comes from the mode solver (the "mode loss") to the loss measured by using the mode monitors in the fdtd simulation (the "fdtd loss"). These losses are plotted as a function of waveguide width. Ideally the curves from both loss calculations should look very similar to each other. Here is the behavior before and after implementing the PML refactor:

Before
image

After
image
(I get even better agreement, almost perfect, if I run with higher grid precision)

As I said before, please let me know what you think. As is, this refactor changes many files and so I would imagine would require quite some review.


ymahlau:
On a side note, I have refactored the place_objects function. The new API is slightly different, please do not be confused. I will add more documentation soon. This was necessary for the current development of a graphical user interface for fdtdx.

src/fdtdx/fdtd/update.py
```diff
@@ -48,34 +51,134 @@ def update_E(
...
+    # Handle periodic boundaries by copying values from opposite sides
     if simulate_boundaries:
+        E = arrays.E
         for boundary in objects.boundary_objects:
-            boundary_states[boundary.name] = boundary.update_E_boundary_state(
-                boundary_state=arrays.boundary_states[boundary.name],
-                H=arrays.H,
-            )
+            if isinstance(boundary, PeriodicBoundary):
+                E = E.at[..., boundary.boundary_slice].set(E[..., boundary.opposite_slice])
```
galcerte: `boundary.boundary_slice` and `boundary.opposite_slice` should actually be `*boundary.boundary_slice` and `*boundary.opposite_slice`, since otherwise I get this exception `raise IndexError("Indexing mode not yet supported. Got unsupported indexer "
IndexError: Indexing mode not yet supported. Got unsupported indexer at position 3: (slice(0, 1, None), slice(0, 200, None), slice(0, 250, None))`

src/fdtdx/fdtd/update.py
```diff
@@ -196,33 +288,134 @@ def update_H(
...
+    # Handle periodic boundaries by copying values from opposite sides
     if simulate_boundaries:
+        H = arrays.H
         for boundary in objects.boundary_objects:
-            boundary_states[boundary.name] = boundary.update_H_boundary_state(
-                boundary_state=arrays.boundary_states[boundary.name],
-                E=arrays.E,
-            )
+            if isinstance(boundary, PeriodicBoundary):
+                H = H.at[..., boundary.boundary_slice].set(H[..., boundary.opposite_slice])
```
galcerte: Same deal as my previous comment in here.

src/fdtdx/fdtd/update.py
```diff
@@ -48,34 +51,134 @@ def update_E(
...
+        psi_Ezy = b_y * psi_Ezy + a_y * DyHx
+
+        psi_E = jnp.stack((psi_Exy, psi_Exz, psi_Eyz, psi_Eyx, psi_Ezx, psi_Ezy), axis=0)
+        arrays = arrays.aset("psi_E", psi_E)
+
+    # Curl equations
+    curl_x = (1.0 / arrays.kappa[1, :, :, :] * DyHz + psi_Exy) - (1.0 / arrays.kappa[2, :, :, :] * DzHy + psi_Exz)
+    curl_y = (1.0 / arrays.kappa[2, :, :, :] * DzHx + psi_Eyz) - (1.0 / arrays.kappa[0, :, :, :] * DxHz + psi_Eyx)
+    curl_z = (1.0 / arrays.kappa[0, :, :, :] * DxHy + psi_Ezx) - (1.0 / arrays.kappa[1, :, :, :] * DyHx + psi_Ezy)
+    curl = jnp.stack((curl_x, curl_y, curl_z), axis=0)
```
ymahlau: Can you refactor all of the update code into separate functions, similar to how it looked before? I would like to prevent having functions that are very long and hard to maintain

src/fdtdx/fdtd/update.py
rachsmith1: I have simply deleted these lines, since the periodic boundary update is already handled by the curl equations, see #193 for more info.

src/fdtdx/fdtd/update.py
rachsmith1: Same as above, see #193.

src/fdtdx/fdtd/update.py
rachsmith1: I have refactored update.py and curl.py so that the functions aren't as long.

rachsmith1:
@ymahlau The one test that is failing is `test_detector_convergence_condition` in test_condition.py. I saw that you commented it out, but it passes locally for me. I am running python 3.13.

rachsmith1:
> I'm testing out this PR further on the example I previously shared on a gist and I'm seeing some weird "dragging" on the fields as radiation propagates on PMLs oriented normally on maximum X axis and minimum Y axis, but not on the other two PML boundaries opposite to these, where the fields are not affected, which is the expected behaviour: https://meep.readthedocs.io/en/latest/FAQ/#why-are-the-fields-not-being-absorbed-by-the-pml.

Hi @galcerte, I did a deep dive into this and I don't think the issue is with the PML per se, but rather with the treatment of the very edges of the simulation regardless of boundary type. You have to make some decision about what to do with the "ghost cells", that is the values for E and H that are just outside of the volume and consequently "don't exist". Right now FDTDx [nominally sets those values to zero](https://github.com/ymahlau/fdtdx/blob/main/src/fdtdx/core/physics/curl.py#L35).

Now that's fine if you have a PML with a normally incident wave since the PML should attenuate the fields to zero by the time they reach the edge of the simulation and there should be no mismatch. However in your case you have the plane source extending into the PML and driving the fields right next to this zero boundary. This creates a mismatch and the "dragging" distortion that you observe. You can see that if you change the polarization direction of the source the walls that are affected switch. 

This can be fixed by changing what you do with the ghost cells. Instead of setting them to zero, you could set them to the last "real" value in the array. Or you could set them according to the first (or higher order) derivative when you look at the field gradient of the subsequent real array values. The choice seems to depend on what exactly you're trying to achieve with the simulation. These are the results I get using the "reflect-odd" padding option, which effectively does the first order linear extrapolation:

```
    # Pad each axis separately based on boundary conditions
    E_pad = E
    for i, periodic in enumerate(periodic_axes):
        # Create padding tuple for current axis
        if i == 0:
            pad_width = ((0, 0), (0, 1), (0, 0), (0, 0))
        elif i == 1:
            pad_width = ((0, 0), (0, 0), (0, 1), (0, 0))
        else:  # i == 2
            pad_width = ((0, 0), (0, 0), (0, 0), (0, 1))
        if periodic:
            E_pad = jnp.pad(E_pad, pad_width, mode="wrap")
        else:
            E_pad = jnp.pad(E_pad, pad_width, mode="reflect", reflect_type="odd")
```
image
image

I don't think this really constitutes a bug or needs another commit. It's more to do with how you formulate the setup of the simulation and what you're going for. For example, this particular issue can be fixed by just not extending the source into the PML boundaries, or changing the PML boundaries on the X and Y axes to be periodic boundaries instead.

rachsmith1:
I have set the initial PML parameters to reasonable values for a given simulation by default. Of course, the user can always specify custom PML parameters if they choose. I have also gone through all of the scripts in the examples folder and tested them, this includes backwards sims and gradient based optimization - all looks fine to me.

@ymahlau @galcerte Please let me know if there is anything else preventing this from being merged.

tests/fdtd/test_update.py
```diff
@@ -142,30 +146,30 @@ def test_basic_update(self, setup):
...
+    #    with patch.object(arrays, "at") as mock_at:
+    #        mock_setter = Mock()
+    #        mock_setter.set.return_value = arrays
+    #        mock_at.return_value = mock_setter
+    #
+    #        update_E(time_step=jnp.array(0), arrays=arrays, objects=objects, config=config, simulate_boundaries=True)
+    #
+    #    # Verify boundary methods were called
+    #    mock_boundary.update_E_boundary_state.assert_called_once()
+    #    mock_boundary.update_E.assert_called_once()
```
ymahlau: I think this test case can be deleted, it did not do much before anyways

tests/fdtd/test_update.py
```diff
@@ -310,30 +323,30 @@ def test_basic_update(self, setup):
...
+    #    with patch.object(arrays, "at") as mock_at:
+    #        mock_setter = Mock()
+    #        mock_setter.set.return_value = arrays
+    #        mock_at.return_value = mock_setter
+    #
+    #        update_H(time_step=jnp.array(0), arrays=arrays, objects=objects, config=config, simulate_boundaries=True)
+    #
+    #    # Verify boundary methods were called
+    #    mock_boundary.update_H_boundary_state.assert_called_once()
+    #    mock_boundary.update_H.assert_called_once()
```
ymahlau: same for this one

tests/fdtd/test_update.py
rachsmith1: Deleted in latest commit

tests/fdtd/test_update.py
rachsmith1: Deleted in latest commit

Issue #196, opened 2025-11-23
Refactor plot_setup


ymahlau:
The plot_setup function (https://github.com/ymahlau/fdtdx/blob/main/src/fdtdx/utils/plot_setup.py) should be refactored and some new functionality should be added. The existing function plots an image of the objects in the scene from three different angles. 

Please add a new function called plot_setup_from_side with the following functionality:
- Same arguments as plot_setup function. rename axs to ax since this only needs a single matplotlib axis.
- The function should take an additional argument viewing_side, which can be 'x', 'y' or 'z'
- The function should take an additional argument exclude_large_object_ratio (float | None, default None), which if given removes all objects that cover more percentage of the image than the threshold determined by this argument (for example with input 1, only remove objects that cover 100% of the image).
- The function should plot one of the three images of the plot_setup function.

Change the plot_setup function in the following way:
- add the exclude_large_object_ratio argument
- instead of the existing code, call the function above three times (with corresponding axs, viewing_angle argument) to plot the three images next to each other

Add test functions, which call the functions above on a simple simulation setup (use real objects, no mocks, with pytest fixture to re-use the whole setup. You can write a fixture function which generates the full object-container and config object). The test cases should save the plots to a directory tests/generated, which may need to be generated in code. add this directory to the .gitignore
- three test cases for plot_setup_from_side, one from each side
- one test case for plot_setup
- one test case for plot_setup with exclude_large_object_ratio, where a large object is excluded.

renaissancenerd:
@ymahlau Yes I can work on it.

PR #197, opened 2025-11-24
refactored plot_setup

renaissancenerd:
refactored plot_setup by adding a helper function plot_setup_from_side and added .gitignore for generated test plots. The test suite for the source code is yet to follow.

### DIFF
--- a/src/fdtdx/objects/object.py
+++ b/src/fdtdx/objects/object.py
@@ -798,6 +798,7 @@ class OrderableObject(SimulationObject):
                     partial_grid_shape=(10, 10, 10),
                     name="large_background",
                 )
+                self.color = "#e0e0e0"
 
         # Add a smaller centered object
         class CenterObject(SimulationObject):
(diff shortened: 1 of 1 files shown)


renaissancenerd:
@ymahlau I have refactored the plot_setup and added its tests as per #196

tests/utils/test_plot_setup.py
```diff
@@ -0,0 +1,250 @@
...
+
+    # Create simulation volume (10 x 10 x 10 µm = 10e-6 meters)
+    volume = SimulationVolume(
+        partial_real_shape=(10e-6, 10e-6, 10e-6),  # 10 µm in each direction
+        name="simulation_volume",
+    )
+
+    # Add a large background object (covers most of the volume)
+    class LargeObject(SimulationObject):
+        def __init__(self):
```
ymahlau: I think it is a bit ugly to define so many new classes in the test cases. The best way to handle this would probably be to define a new class in fdtdx.objects.object.py which is called EmptySimulationObject as a subclass of OrderableObject. Then you can use this class in the test cases

ymahlau:
Apart from the minor comments above, this looks really good! Many thanks and well done!

Issue #201, opened 2025-11-30
Add plot_material function


ymahlau:
Please add a plot_material and plot_material_from_side function similar to the plot_setup /plot_setup_from_side functions. Instead of the simulation object, these new functions should plot the material distribution at a specific 2d "slice" through the 3d simulation volume. The API should be:

```
MaterialType = Literal["permittivity", "permeability"]

def plot_material_from_side(
    config: SimulationConfig,
    arrays: ArrayContainer,
    viewing_side: Literal["x", "y", "z"],
    filename: str | Path | None = None,
    ax: Any | None = None,
    plot_legend: bool = True,
    position: float = 0,  # this is the position of the slice. Zero means at center, 1e-6 would mean center+1µm
    type: str = "permittivity",
) -> Figure:

def plot_material(
    config: SimulationConfig,
    arrays: ArrayContainer,
    viewing_side: Literal["x", "y", "z"],
    filename: str | Path | None = None,
    ax: Any | None = None,
    plot_legend: bool = True,
    positions: tuple[float, float, float] = 0,  # position of slices in x, y, z
    type: str = "permittivity",
) -> Figure:
```

Depending on the type argument, either the material from arrays.permittivity or arrays.permeability should be plotted. 

You can calculate the 2d slice to be plotted from the 3d array by using the config.resolution. So round(position / config.resolution) gives you the number of array indexes you need to add/subtract from the center of the 3d array.

please add test cases similar to the test cases from the plot_setup functions. Use UniformMaterialObject, Cylinder and Sphere objects in the tests.

ymahlau:
@renaissancenerd could you tackle this issue please?

renaissancenerd:
> [@renaissancenerd](https://github.com/renaissancenerd) could you tackle this issue please?

Yes I can work on it.

PR #202, opened 2025-12-01
added plot_material and its helper function plot_material_from_side

renaissancenerd:
For #201 , added plot_material and its helper function plot_material_from_side. The test suite for the said functions will follow.


ymahlau:
Awesome! thanks for finishing this so quickly!

Issue #75, opened 2025-06-14
Non-isotropic materials


ymahlau:
Implement non-isotropic materials by changing the shape of the permittivity / permeability array. They should have shape (3, Nx, Ny, Nz) for non-isotropic and (1, Nx, Ny, Nz) for isotropic materials. Also change the API to make specification of these materials easy.

ymahlau:
See #68

bruxillensis:
A question regarding this implementation I'd like some feedback on:

fdtdx.StandardToCustomRange, fdtdx.StandardToInversePermittivityRange, and fdtdx.StandardToPlusOneMinusOneRange are all linear maps of the permitivity. For an anisotropic material, what would make the most sense for these functions? Some info on how this is used would be helpful as I couldn't find any examples. I can think of a few options depending on the use case:

1. Return a 3x1 vector, where the 3 dimensions are kept separate. Normalization for each axis is separately calculated
2. Return a 3x1 vector, same as above. Normalization is based on the max/min of the entire tensor.
4. Return a 1x1 map based on the max/min of the entire tensor.

PR #204, opened 2025-12-05
Implemented anisotropic materials

bruxillensis:
Implemented linear diagnonal anisotropic materials (no cross-diagonal components) #75.

Example simulation simulate_gaussian_source_anisotropic.py added.
Energy and mode test cases with anisotropic materials added.
Added test cases for anisotropic material.
Refactored code to handle (3, nx, ny, nz) material matrices.

### DIFF
--- a/src/fdtdx/core/physics/modes.py
+++ b/src/fdtdx/core/physics/modes.py
@@ -185,6 +185,26 @@ def compute_mode(
         indices=0,
         axis=propagation_axis + 1,
     )
+
+    # Rotate permittivity components to match tidy3d coordinate system
+    # tidy3d assumes propagation along z, so we need to map physical axes to tidy3d axes:
+    # - tidy3d x → first transverse axis
+    # - tidy3d y → second transverse axis
+    # - tidy3d z → propagation axis
+    if propagation_axis == 0:
+        # propagation along x: tidy3d (x,y,z) → physical (y,z,x)
+        perm_idx = [1, 2, 0]
+    elif propagation_axis == 1:
+        # propagation along y: tidy3d (x,y,z) → physical (x,z,y)
+        perm_idx = [0, 2, 1]
+    else:  # propagation_axis == 2
+        # propagation along z: tidy3d (x,y,z) → physical (x,y,z)
+        perm_idx = [0, 1, 2]
+
+    # Only apply rotation if anisotropic (3 components)
+    if permittivity_squeezed.shape[0] == 3:
+        permittivity_squeezed = permittivity_squeezed[jnp.array(perm_idx), :, :]
+
     result_shape_dtype = (
         jnp.zeros((3, *permittivity_squeezed.shape[1:]), dtype=jnp.complex64),
         jnp.zeros((3, *permittivity_squeezed.shape[1:]), dtype=jnp.complex64),
@@ -198,6 +218,9 @@ def compute_mode(
             indices=0,
             axis=propagation_axis + 1,
         )
+        # Apply same rotation to permeability if anisotropic
+        if permeability_squeezed.shape[0] == 3:
+            permeability_squeezed = permeability_squeezed[jnp.array(perm_idx), :, :]
     else:  # float
         permeability_squeezed = permeabilities
 
--- a/tests/core/physics/test_modes.py
+++ b/tests/core/physics/test_modes.py
@@ -244,6 +244,246 @@ class TestComputeMode:
             assert E.shape == (3, 3, 1, 4)  # Note: axis handling in the function
 
 
+class TestAnisotropicModeComputation:
+    """Test anisotropic material handling in compute_mode."""
+
+    @patch("fdtdx.core.physics.modes.tidy3d_mode_computation_wrapper")
+    @patch("fdtdx.core.physics.modes.normalize_by_poynting_flux")
+    def test_anisotropic_permittivity_rotation_propagation_axis_0(self, mock_normalize, mock_tidy3d_wrapper):
+        """Test that anisotropic permittivity is rotated correctly for propagation along axis 0.
+
+        When propagation is along x (axis 0), tidy3d (x,y,z) maps to physical (y,z,x).
+        So physical permittivity (eps_x, eps_y, eps_z) should become tidy3d (eps_y, eps_z, eps_x).
+        """
+        mock_mode = ModeTupleType(
+            neff=1.5 + 0.1j,
+            Ex=np.ones((5, 6), dtype=np.complex64),
+            Ey=np.ones((5, 6), dtype=np.complex64),
+            Ez=np.ones((5, 6), dtype=np.complex64),
+            Hx=np.ones((5, 6), dtype=np.complex64),
+            Hy=np.ones((5, 6), dtype=np.complex64),
+            Hz=np.ones((5, 6), dtype=np.complex64),
+        )
+        mock_tidy3d_wrapper.return_value = [mock_mode]
+        mock_normalize.return_value = (
+            jnp.ones((3, 1, 5, 6), dtype=jnp.complex64),
+            jnp.ones((3, 1, 5, 6), dtype=jnp.complex64),
+        )
+
+        # Anisotropic permittivity with distinct values: eps_x=2, eps_y=3, eps_z=4
+        # Shape: (3, 1, 5, 6) - propagation along axis 0
+        inv_permittivities = jnp.zeros((3, 1, 5, 6))
+        inv_permittivities = inv_permittivities.at[0].set(1 / 2.0)  # eps_x = 2
+        inv_permittivities = inv_permittivities.at[1].set(1 / 3.0)  # eps_y = 3
+        inv_permittivities = inv_permittivities.at[2].set(1 / 4.0)  # eps_z = 4
+
+        compute_mode(
+            frequency=2e14,
+            inv_permittivities=inv_permittivities,
+            inv_permeabilities=1.0,
+            resolution=1e-8,
+            direction="+",
+        )
+
+        # Check what was passed to tidy3d_wrapper
+        call_args = mock_tidy3d_wrapper.call_args
+        perm_passed = call_args.kwargs["permittivity_cross_section"]
+
+        # For propagation along axis 0: perm_idx = [1, 2, 0]
+        # So tidy3d should receive (eps_y, eps_z, eps_x) = (3, 4, 2)
+        assert perm_passed.shape[0] == 3
+        assert np.allclose(perm_passed[0], 3.0)  # tidy3d x component = physical y
+        assert np.allclose(perm_passed[1], 4.0)  # tidy3d y component = physical z
+        assert np.allclose(perm_passed[2], 2.0)  # tidy3d z component = physical x
+
+    @patch("fdtdx.core.physics.modes.tidy3d_mode_computation_wrapper")
+    @patch("fdtdx.core.physics.modes.normalize_by_poynting_flux")
+    def test_anisotropic_permittivity_rotation_propagation_axis_1(self, mock_normalize, mock_tidy3d_wrapper):
+        """Test that anisotropic permittivity is rotated correctly for propagation along axis 1.
+
+        When propagation is along y (axis 1), tidy3d (x,y,z) maps to physical (x,z,y).
+        So physical permittivity (eps_x, eps_y, eps_z) should become tidy3d (eps_x, eps_z, eps_y).
+        """
+        mock_mode = ModeTupleType(
+            neff=1.5 + 0.1j,
+            Ex=np.ones((5, 6), dtype=np.complex64),
+            Ey=np.ones((5, 6), dtype=np.complex64),
+            Ez=np.ones((5, 6), dtype=np.complex64),
+            Hx=np.ones((5, 6), dtype=np.complex64),
+            Hy=np.ones((5, 6), dtype=np.complex64),
+            Hz=np.ones((5, 6), dtype=np.complex64),
+        )
+        mock_tidy3d_wrapper.return_value = [mock_mode]
+        mock_normalize.return_value = (
+            jnp.ones((3, 5, 1, 6), dtype=jnp.complex64),
+            jnp.ones((3, 5, 1, 6), dtype=jnp.complex64),
+        )
+
+        # Anisotropic permittivity with distinct values: eps_x=2, eps_y=3, eps_z=4
+        # Shape: (3, 5, 1, 6) - propagation along axis 1
+        inv_permittivities = jnp.zeros((3, 5, 1, 6))
+        inv_permittivities = inv_permittivities.at[0].set(1 / 2.0)  # eps_x = 2
+        inv_permittivities = inv_permittivities.at[1].set(1 / 3.0)  # eps_y = 3
+        inv_permittivities = inv_permittivities.at[2].set(1 / 4.0)  # eps_z = 4
+
+        compute_mode(
+            frequency=2e14,
+            inv_permittivities=inv_permittivities,
+            inv_permeabilities=1.0,
+            resolution=1e-8,
+            direction="+",
+        )
+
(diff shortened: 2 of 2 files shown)


ymahlau:
Hi @bruxillensis, many thanks for this contribution! It is awesome to see this feature realized! Since this is a bit of a bigger PR and due to my current internship I have only little time it might take me a few days to review to this PR. 

@l-berg Could you also take a look at this PR and give it a review?

l-berg:
Hi @bruxillensis Thanks for putting in the effort to realize this feature. In your current implementation, we are always paying for the larger 3-dimensional inv_permittivities and inv_permeabilities. If all materials are isotropic, keeping these 1-dimensional (e.g., via `shape=(1, nx, ny, nz)` instead of `(3, nx, ny, nz)`) would require ~5% less memory and be a little faster.

Instead of strictly converting to tuples at the material stage, we could handle this at the initialization stage and only use 3-dimensional permittivities/permeabilities for the simulation buffers if one or more objects actually use an anisotropic material.

I think this would be a better approach. What do you think?

src/fdtdx/objects/device/parameters/continuous.py
```diff
@@ -29,10 +29,14 @@ def __call__(
     ) -> dict[str, jax.Array]:
         del kwargs
         # determine minimum and maximum allowed permittivity
+        # For anisotropic materials, use average of components
         max_inv_perm, min_inv_perm = -math.inf, math.inf
```
l-berg: I do not get why you use the average here. The goal of this `ParameterTransform` is to interpolate between permittivities using values between 0 and 1. For 2 anisotropic materials, we can easily interpolate between them so that 0->material0 and 1->material1, and 0.5->0.5*(material0.permittivity+material1.permittivity) etc.. It is not obvious to me how this would work for more than 2 anisotropic materials, so maybe we should just add an assert for now which forbids this case.

bruxillensis: This is similar to the question I asked in #75. @ymahlau suggested to return a 3-tuple of each mapped component. I could see how this might work, with the interpolation being component wise. Changing it based on his answer is something I forgot to do.

src/fdtdx/fdtd/initialization.py
```diff
@@ -312,41 +322,78 @@ def _init_arrays(
     info = {}
     for o in sorted_obj:
         if isinstance(o, UniformMaterialObject):
-            inv_permittivities = inv_permittivities.at[*o.grid_slice].set(1 / o.material.permittivity)
+            # Material properties are tuples (εx, εy, εz)
+            # Arrays have shape (3, Nx, Ny, Nz) where first axis is component
+            # Set each component: inv_permittivities[i, x, y, z] = 1/ε_i
+            for i in range(3):
```
l-berg: Using for loops here is not ideal. We should do something like
```
inv_obj_permittivity = (1 / jnp.array(o.material.permittivity))[:, None, None, None]
inv_permittivities = inv_permittivities.at[:, *o.grid_slice].set(inv_obj_permittivity)
```
Maybe we should even expose `o.material.permittivity` as an array instead of tuple in the first place.

examples/simulate_gaussian_source_anisotropic.py
```diff
@@ -0,0 +1,223 @@
+"""
+This script exemplifies a simple simulation in fdtdx with an anisotropic boundary. In this simulation, a simulation volume with PML
+boundaries is set up. At the top of the simulation, a source is placed, which outputs a continous light wave of
+1550nm wavelength. At the bottom of the simulation, a block of material with an anistropic refractive index of (2.0, 1.0, 2.0) for the
+x, y and z axes is placed. On line 22 the polarization of the gaussian source can be adjusted. If the source is X polarized, the resulting
```
l-berg: Great idea to have an anisotropic example. "line 22" in the initial doc-string is outdated.

examples/simulate_gaussian_source_anisotropic.py
```diff
@@ -0,0 +1,223 @@
...
+        final_state = fdtdx.run_fdtd(
+            arrays=arrays,
+            objects=new_objects,
+            config=config,
+            key=key,
+        )
+        _, arrays = final_state
+
+        # Run backward simulation (for gradients or adjoint fields)
+        _, arrays = fdtdx.full_backward(
```
l-berg, line 183: The backward video detector part was removed in order to speed up the script? As a result, I think currently the results of the backward simulation are not used anywhere.

src/fdtdx/core/physics/modes.py
```diff
@@ -262,32 +262,44 @@ def tidy3d_mode_computation_wrapper(
         track_freq="central",
         group_index_step=False,
     )
-    od = np.zeros_like(permittivity_cross_section)
+    od = np.zeros_like(permittivity_cross_section[1, :, :])
     eps_cross = [
-        permittivity_cross_section,
+        jnp.squeeze(permittivity_cross_section[0, :, :]),
```
l-berg: What does `jnp.squeeze` accomplish here? Isn't `permittivity_cross_section[0, :, :]` a 2d array already?

src/fdtdx/objects/detectors/energy.py
```diff
@@ -65,9 +65,9 @@ def update(
     ) -> DetectorState:
         cur_E = E[:, *self.grid_slice]
         cur_H = H[:, *self.grid_slice]
-        cur_inv_permittivity = inv_permittivity[self.grid_slice]
+        cur_inv_permittivity = inv_permittivity[:, *self.grid_slice]
         if isinstance(inv_permeability, jax.Array) and inv_permeability.ndim > 0:
-            cur_inv_permeability = inv_permeability[self.grid_slice]
+            cur_inv_permeability = inv_permeability[3, *self.grid_slice]
```
rachsmith1: This should be `cur_inv_permeability = inv_permeability[:, *self.grid_slice]`, no?

rachsmith1:
> If I run the script without my PR (commit [9f6d0ef](https://github.com/ymahlau/fdtdx/commit/9f6d0ef138a2b56f6630c2ac497204ee294345e6)) I get slightly different results for the fdtd mode loss than when using my current PR. The mode solver results stay the exact same. The difference is small (1E-6) but I was trying to find the source of the change in case something had been missed in my changes.

@bruxillensis I think this is some kind of floating point issue. I couldn't find anything obvious in your code that would cause this. However, if I run the simulations with `dtype=jnp.float64` for both the main branch and this PR, then the values for `fdtd_loss` are the same within 1e-12.

examples/simulate_gaussian_source.py
```diff
@@ -68,7 +68,7 @@ def main():
     object_list.append(volume)
 
     # Choose boundary type: periodic or PML (absorbing)
-    periodic = True
+    periodic = False
```
ymahlau: As a minor comment, please leave the original example script as is

src/fdtdx/core/grid.py
```diff
@@ -71,7 +80,18 @@ def calculate_time_offset_yee(
...
     # adjust speed for material and calculate time offset
-    refractive_idx = 1 / jnp.sqrt(inv_permittivities * inv_permeabilities)
+    # For component-wise materials, use average of components for refractive index
+    if inv_permittivities.ndim == 4:
+        inv_perm_avg = jnp.mean(inv_permittivities, axis=0)
+    else:
+        inv_perm_avg = inv_permittivities
+
+    if isinstance(inv_permeabilities, jax.Array) and inv_permeabilities.ndim == 4:
+        inv_perm_avg_perm = jnp.mean(inv_permeabilities, axis=0)
```
ymahlau: I think this should be the dot product between wave_vector and permittivity / permeability to be entirely correct, right? The mean is only correct if the wave_vector is at 45 degree angle. I would guess that this does not make a big difference, but might lead to backwards inserted energy in the TFSF-Source.

src/fdtdx/fdtd/fdtd.py
```diff
@@ -63,10 +63,13 @@ def reversible_fdtd(
...
         arr: ArrayContainer,
     ) -> SimulationState:
+        def cond(s):
+            return config.time_steps_total > s[0]
+
         state = (jnp.asarray(0, dtype=jnp.int32), arr)
         state = eqxi.while_loop(
             max_steps=config.time_steps_total,
-            cond_fun=lambda s: config.time_steps_total > s[0],
+            cond_fun=cond,
```
rachsmith1: Why was this changed?

examples/simulate_gaussian_source_anisotropic.py
bruxillensis: Yes, since it was demonstrated in the isotropic script I removed it for speed.

src/fdtdx/core/grid.py
bruxillensis: You are right the mean is very wrong. However, I think k dot esp or mu would only be if the injection angle is normal to the source plane... Since the gaussian and plane wave sources can have arbitrary injection angles, we need to use the polarization vector to calculate the correct esp and mu. Is this correct?

ymahlau: ah yes, you are right that things actually get more complicated in anisotropic media. I think in that case only two distinct polarizations are allowed that experience different refractive indices. 

After a brief research this seems to be a good ressource: https://www.taylorfrancis.com/chapters/mono/10.1201/9780429425356-6/light-propagation-anisotropic-media-myeongkyu-lee?context=ubx&refId=e0c1b513-352b-4c05-afae-289dda1c23bb

We could also postpone this problem to a different PR if you don't immediately need a source inside an anisotropic material. In that case you could add a check that throws an error here if an anisotropic material is provided.

rachsmith1: I just wanted to mention that I am interested in sources inside of anisotropic materials. I haven't looked into the math of this yet but will when I have time.

bruxillensis: Thanks @ymahlau and @rachsmith1! I believe the correct result would follow from the functions treatment of a mode source, by computing the effective index to get the phase velocity, $n_{eff}=p_x^2\epsilon_x+p_y^2\epsilon_y+p_z^2\epsilon_z$ where $\hat{p}$ is the polarization vector. Similarly for $\mu_{eff}$.

Following from the energy density $E=\frac{1}{2}E\cdot D=\frac{1}{2}\epsilon_{eff}E_0^2$, where $E=E_0\hat{p}$, $D=\epsilon E_0\hat{p}$ the effective index would be $\hat{p}\cdot\epsilon\cdot\hat{p}$

Not positive though, I havent found an FDTD textbook that discusses it specifically. In the meantime, I've implemented the above, since it simplifies to the correct result for the isotropic case, and I will throw an error until one of us can verify/correct it.

src/fdtdx/fdtd/fdtd.py
bruxillensis: Left from debugging by accident. Corrected with 4e7ea16.

src/fdtdx/core/grid.py
bruxillensis: Implemented with 4e7ea16.

examples/simulate_gaussian_source.py
bruxillensis: Left from debugging by accident. Corrected with 4e7ea16.

src/fdtdx/objects/detectors/energy.py
bruxillensis: Corrected with 4e7ea16.

src/fdtdx/core/physics/modes.py
bruxillensis: Corrected with 4e7ea16.

examples/simulate_gaussian_source_anisotropic.py
bruxillensis: Corrected with 4e7ea16.

src/fdtdx/fdtd/initialization.py
bruxillensis: Tuples were kept but converted to jnp arrays in initialization.py with commit 4e7ea16.

src/fdtdx/objects/device/parameters/continuous.py
bruxillensis: Corrected with 4e7ea16 to implement @ymahlau suggestion in #75.

src/fdtdx/core/physics/modes.py
```diff
@@ -169,17 +177,17 @@ def mode_helper(permittivity, permeability):
...
-        axis=propagation_axis,
+        axis=propagation_axis + 1,
     )
     result_shape_dtype = (
-        jnp.zeros((3, *permittivity_squeezed.shape), dtype=jnp.complex64),
-        jnp.zeros((3, *permittivity_squeezed.shape), dtype=jnp.complex64),
+        jnp.zeros((3, *permittivity_squeezed.shape[1:]), dtype=jnp.complex64),
+        jnp.zeros((3, *permittivity_squeezed.shape[1:]), dtype=jnp.complex64),
         jnp.zeros(shape=(), dtype=jnp.complex64),
     )
```
rachsmith1, line 213: @bruxillensis Great work so far. This is looking very good. Noticed this bug. Tidy3D assumes propagation in the z direction, so before passing the permittivity/permeability to the mode solver they need to be rotated to match Tidy3D's coordinate system. Something like:

```
    # Rotate permittivity components to match tidy3d coordinate system
    # tidy3d assumes propagation along z, so we need to map physical axes to tidy3d axes:
    # - tidy3d x → first transverse axis
    # - tidy3d y → second transverse axis
    # - tidy3d z → propagation axis
    if propagation_axis == 0:
        # propagation along x: tidy3d (x,y,z) → physical (y,z,x)
        perm_idx = [1, 2, 0]
    elif propagation_axis == 1:
        # propagation along y: tidy3d (x,y,z) → physical (x,z,y)
        perm_idx = [0, 2, 1]
    else:  # propagation_axis == 2
        # propagation along z: tidy3d (x,y,z) → physical (x,y,z)
        perm_idx = [0, 1, 2]

    # Only apply rotation if anisotropic (3 components)
    if permittivity_squeezed.shape[0] == 3:
        permittivity_squeezed = permittivity_squeezed[jnp.array(perm_idx), :, :]

    permeabilities = 1 / inv_permeabilities
    if isinstance(inv_permeabilities, jax.Array) and inv_permeabilities.ndim > 0:
        permeability_squeezed = jnp.take(
            permeabilities,
            indices=0,
            axis=propagation_axis + 1,
        )
        # Apply same rotation to permeability if anisotropic
        if permeability_squeezed.shape[0] == 3:
            permeability_squeezed = permeability_squeezed[jnp.array(perm_idx), :, :]
    else:  # float
        permeability_squeezed = permeabilities
```

src/fdtdx/core/grid.py
```diff
@@ -79,21 +80,57 @@ def calculate_time_offset_yee(
...
-
-    if isinstance(inv_permeabilities, jax.Array) and inv_permeabilities.ndim == 4:
-        inv_perm_avg_perm = jnp.mean(inv_permeabilities, axis=0)
+    if effective_index is not None:
+        refractive_idx = effective_index * jnp.ones(spatial_shape)
     else:
-        inv_perm_avg_perm = inv_permeabilities
+        # adjust speed for material and calculate time offset
+        # inv_eps_eff = |p_x|^2 * inv_eps_x + |p_y|^2 * inv_eps_y + |p_z|^2 * inv_eps_z
+        if inv_permittivities.ndim == 4:
```
ymahlau, line 88: For now it is okay to raise an exception here since apply is only called once at the beginning of the simulation. Therefore the debug_callback should not lead to efficiency decreases

src/fdtdx/core/physics/modes.py
bruxillensis: Ouhh yes. Thanks for catching this.

bruxillensis: Fixed with commit 91b2cd9.

PR #206, opened 2025-12-06
added documentation for placeholder attributes

renaissancenerd:


ymahlau:
Please add the descriptions directly above the attribute definitions. The docstring itself should not be changed. The format should be similar to the existing attribute descriptions like here: https://github.com/ymahlau/fdtdx/blob/9f6d0ef138a2b56f6630c2ac497204ee294345e6/src/fdtdx/objects/static_material/sphere.py#L21

ymahlau:
In case there are inherited attributes, for example I think color is sometimes inherited and overriden in the subclasses, then the description should be the same as in the superclass, only the default value should be changed

ymahlau:
And @renaissancenerd please make an entry in the changelog.md for every PR you make from now on. I want to have a well compiled list of changes that we can list in the releases of fdtdx.

PR #209, opened 2025-12-08
added plot_field_slice and its helper function plot_field_slice_component

renaissancenerd:
added plot_field_slice and its helper function plot_field_slice_component. Test suite will follow.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,8 @@
 
 # Unpublished
 ## Added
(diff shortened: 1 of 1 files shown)


ymahlau:
@renaissancenerd you should add the new points in the changelog under unpublished -> added, not v0.5 -> added. v0.5 is already released and your new changes will be released with the next version

Issue #191, opened 2025-11-18
Unitful function for jnp.where / np.where


ymahlau:
There is already some code implemented for the where function, but it needs to be expanded: https://github.com/ymahlau/fdtdx/blob/656480789532b1c70ebb1c52c878e122e663f918/src/fdtdx/functional/numpy.py#L1408

- There should be an overload for all combinations of jax.Array, np.ndarray and Unitful for the x, y, and condition argument. The condition cannot be a Unitful with non-empty unit. 
- You do not need to write down all 27 combinations, but you can simplify a little: 
   - if at least one of the arguments is a Unitful, the result should be a unitful as well, so you can merge both the jax.Array and np.ndarray as ArrayLike.
   - if no argument is Unitful and at least one is a jax.Array, the result should be a jax.Array. Therefore you can merge for the other arguments jax.Array and np.ndarray as ArrayLike.
- if at least one of the inputs is a unitful, you should differentiate if np.where or jnp.where should be executed. If one of the inputs is a jax.Array, then use jnp.where, and use np.where only when all inputs are numpy arrays or Unitfuls with numpy arrays inside. 
- Write test cases covering all lines of codes you write

PR #210, opened 2025-12-11
improved unitful where

sIHURs:
Imprvoed unitful where function, tackles issue https://github.com/ymahlau/fdtdx/issues/191

### DIFF
--- a/src/fdtdx/functional/numpy.py
+++ b/src/fdtdx/functional/numpy.py
@@ -1453,13 +1453,8 @@ def where(
 
     c = condition if isinstance(condition, Unitful) else Unitful(val=condition, unit=EMPTY_UNIT)
     new_x = x if isinstance(x, Unitful) else Unitful(val=x, unit=EMPTY_UNIT)
-    new_y = (
-        y
-        if isinstance(x, Unitful)
-        else Unitful(val=y.val, unit=Unit(scale=y.unit.scale, dim={}), static_arr=y.static_arr)
-    )
 
-    return where(c, new_x, new_y, *args, **kwargs)
+    return where(c, new_x, y, *args, **kwargs)
 
 
 @overload
@@ -1475,13 +1470,8 @@ def where(
 
     c = condition if isinstance(condition, Unitful) else Unitful(val=condition, unit=EMPTY_UNIT)
     new_y = y if isinstance(y, Unitful) else Unitful(val=y, unit=EMPTY_UNIT)
-    new_x = (
-        x
-        if isinstance(y, Unitful)
-        else Unitful(val=x.val, unit=Unit(scale=x.unit.scale, dim={}), static_arr=x.static_arr)
-    )
 
-    return where(c, new_x, new_y, *args, **kwargs)
+    return where(c, x, new_y, *args, **kwargs)
 
 
 @overload
--- a/tests/functional/test_numpy.py
+++ b/tests/functional/test_numpy.py
@@ -1155,11 +1155,9 @@ def test_where_unitful_jax_backend():
 
 def test_where_numpy_and_unitful():
     """Test where with NumPy condition/x and Unitful y"""
-    m = Unit(scale=0, dim={SI.m: 1})
-
     condition = np.array([True, False, True])
     x = np.array([1.0, 2.0, 3.0])
-    y = Unitful(val=np.array([10.0, 20.0, 30.0]), unit=m)
+    y = Unitful(val=np.array([10.0, 20.0, 30.0]), unit=EMPTY_UNIT)
 
     result = where(condition, x, y)
 
@@ -1171,11 +1169,9 @@ def test_where_numpy_and_unitful():
 
 def test_where_jax_and_unitful():
     """Test where with JAX condition/x and Unitful y"""
-    m = Unit(scale=0, dim={SI.m: 1})
-
     condition = jnp.array([True, False, True])
     x = jnp.array([1.0, 2.0, 3.0])
-    y = Unitful(val=np.array([10.0, 20.0, 30.0]), unit=m)
+    y = Unitful(val=np.array([10.0, 20.0, 30.0]), unit=EMPTY_UNIT)
 
     result = where(condition, x, y)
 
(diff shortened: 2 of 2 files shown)


src/fdtdx/functional/numpy.py
```diff
@@ -1416,7 +1416,14 @@ def where(
     assert condition.unit.dim == {} and condition.unit.scale == 0, f"Invalid condition input: {condition}"
     x_align, y_align = align_scales(x, y)
     c_val = condition.val
-    new_val = jnp._orig_where(c_val, x_align.val, y_align.val, *args, **kwargs)  # type: ignore
+
+    if any(isinstance(v, jax.Array) for v in [x_align.val, y_align.val, c_val]):
+        new_val = jnp._orig_where(c_val, x_align.val, y_align.val, *args, **kwargs)  # type: ignore
+    elif all(isinstance(v, np.ndarray) for v in (x_align.val, y_align.val, c_val)):
```
ymahlau: I believe this should be als 'any', not 'all', since there may be python floats / ints / bool etc as input

src/fdtdx/functional/numpy.py
```diff
@@ -1425,42 +1432,53 @@ def where(
...
             new_static_arr = np.where(c_arr, x_arr, y_arr, *args, **kwargs)
-    return Unitful(val=new_val, unit=x.unit, static_arr=new_static_arr)
+    return Unitful(val=new_val, unit=x_align.unit, static_arr=new_static_arr)
 
 
 @overload
 def where(
-    condition: jax.Array | Unitful,
+    condition: ArrayLike | Unitful,
     x: ArrayLike,
```
ymahlau: Here x can also be ArrayLike | Unitful. The pattern should be that if at least one input is unitful, the all other other inputs can be whatever and the output is unitful

src/fdtdx/functional/numpy.py
```diff
@@ -1425,42 +1432,53 @@ def where(
...
-    return where(c, Unitful(val=x, unit=EMPTY_UNIT), y, *args, **kwargs)
+    return where(c, Unitful(val=x, unit=EMPTY_UNIT), y_non_dim, *args, **kwargs)
 
 
 @overload
 def where(
-    condition: jax.Array | Unitful,
+    condition: ArrayLike | Unitful,
     x: Unitful,
     y: ArrayLike,
```
ymahlau: Similarly, y can be ArrayLike | Unitful here

src/fdtdx/functional/numpy.py
```diff
@@ -1425,42 +1432,53 @@ def where(
...
+    x_non_dim = Unitful(val=x.val, unit=new_unit)
     c = condition if isinstance(condition, Unitful) else Unitful(val=condition, unit=EMPTY_UNIT)
-    return where(c, x, Unitful(val=y, unit=EMPTY_UNIT), *args, **kwargs)
+    return where(c, x_non_dim, Unitful(val=y, unit=EMPTY_UNIT), *args, **kwargs)
 
 
 @overload
 def where(
-    condition: jax.Array,
+    condition: ArrayLike,
```
ymahlau: Here we need more fine-grained functions. Please add four functions for jax.Array output similar to the unitful functions above (one functions with all inputs jax.Array and three functions where one of the inputs is definitely jax.Array and the other inputs can be whatever). Then add one function for numpy / python input that returns a numpy array.

sIHURs: thanks for the comments. I split the function into more fine-grained overloads. But I didn’t add a separate overload for the case where all inputs are jax.Array, because that caused overlapping overload errors. The current three functions return jax.Array and the one returns np.ndarray  I think cover inputs type.

src/fdtdx/functional/numpy.py
```diff
@@ -1425,31 +1437,62 @@ def where(
...
+    if isinstance(condition, Unitful) and isinstance(x, Unitful):
+        return where(condition, x, y, *args, **kwargs)
+
     c = condition if isinstance(condition, Unitful) else Unitful(val=condition, unit=EMPTY_UNIT)
-    return where(c, Unitful(val=x, unit=EMPTY_UNIT), y, *args, **kwargs)
+    new_x = x if isinstance(x, Unitful) else Unitful(val=x, unit=EMPTY_UNIT)
+    new_y = (
+        y
+        if isinstance(x, Unitful)
+        else Unitful(val=y.val, unit=Unit(scale=y.unit.scale, dim={}), static_arr=y.static_arr)
```
ymahlau: this if else is a bit superficial and can actually lead to bugs. The correct way would be to just use y always I think

src/fdtdx/functional/numpy.py
```diff
@@ -1425,31 +1437,62 @@ def where(
...
+    if isinstance(y, Unitful) and isinstance(condition, Unitful):
+        return where(condition, x, y, *args, **kwargs)
+
     c = condition if isinstance(condition, Unitful) else Unitful(val=condition, unit=EMPTY_UNIT)
-    return where(c, x, Unitful(val=y, unit=EMPTY_UNIT), *args, **kwargs)
+    new_y = y if isinstance(y, Unitful) else Unitful(val=y, unit=EMPTY_UNIT)
+    new_x = (
+        x
+        if isinstance(y, Unitful)
+        else Unitful(val=x.val, unit=Unit(scale=x.unit.scale, dim={}), static_arr=x.static_arr)
```
ymahlau: same here

tests/functional/test_numpy.py
```diff
@@ -1113,3 +1115,137 @@ def test_arctanh_unitful_with_jax_array():
...
+        if is_currently_compiling() and not unitful.STATIC_OPTIM_STOP_FLAG:
+            assert out.static_arr is not None
+        return out
+
+    jitted_fn = jax.jit(fn)
+    result = jitted_fn(c, x, y)
+
+    expected = np.where(np.array([True, False, True]), np.array([1.0, 2.0, 3.0]), np.array([10.0, 20.0, 30.0]))
+
+    assert jnp.allclose(result.value(), jnp.array(expected))
```
ymahlau, line 1247: Can you add another test case which tetsts that invalid inputs also lead to throwing an error? For example, test that an exception is thrown if x and y have different units

src/fdtdx/functional/numpy.py
sIHURs: There’s a question here: we need to keep the dims of x and y consistent.
When an ArrayLike input (for example x) is promoted to Unitful, should we assign x the unit.dim of the other Unitful input (for example y), or should we instead strip the unit.dim from y?

ymahlau: This is exactly what I am talking about. If the unit.dim are different then an Exception should be thrown. I believe you can just let the units as is and let the error handling in the other unitful where definition take care of it

Issue #87, opened 2025-06-27
remove multiple wave_character in PhasorDetector


ymahlau:
The PhasorDetector has multiple wave_characters in the signature, but currently only supports a single frequency. Change the signature and shape computation.

ymahlau:
Or even better, implement measuring multiple frequencies.

PR #216, opened 2026-01-06
updated PhasorDetector for multiple frequencies

renaissancenerd:
For #87, the implementation now properly handles multiple frequencies simultaneously through vectorized operations, computing the phasor for each frequency-component-position

### DIFF
--- a/src/fdtdx/objects/detectors/phasor.py
+++ b/src/fdtdx/objects/detectors/phasor.py
@@ -88,18 +88,24 @@ class PhasorDetector(Detector):
         if "Hz" in self.components:
             fields.append(H[2])
 
-        EH = jnp.stack(fields, axis=0)
+        EH = jnp.stack(fields, axis=0)  # Shape: (num_components, *grid_shape)
 
         # Vectorized phasor calculation for all frequencies
-        phase_angles = self._angular_frequencies * time_passed  # Shape: (num_freqs,)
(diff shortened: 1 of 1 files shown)


src/fdtdx/objects/detectors/phasor.py
```diff
@@ -88,18 +88,24 @@ def update(
...
+        EH = jnp.stack(fields, axis=0)  # Shape: (num_components, *grid_shape)
 
         # Vectorized phasor calculation for all frequencies
-        phase_angles = self._angular_frequencies * time_passed  # Shape: (num_freqs,)
-        phasors = jnp.exp(1j * phase_angles)  # Shape: (num_freqs,)
-        # Reshape phasors to (num_freqs, 1, 1, 1, 1) for proper broadcasting with EH (num_components, x, y, z)
-        phasors = phasors.reshape((len(self._angular_frequencies),) + (1,) * EH.ndim)
-        new_phasors = EH * phasors * static_scale  # Shape: (num_freqs, num_components, *grid_shape)
+        phase_angles = self._angular_frequencies[:, None] * time_passed  # Shape: (num_freqs, 1)
+        phasors = jnp.exp(1j * phase_angles)  # Shape: (num_freqs, 1)
```
ymahlau, line 95: I think it is a bit unnecessary to add the second dimension of size 1 here if you reshape below anyways.

Issue #220, opened 2026-01-12
Detector plotting issue


rachsmith1:
Shouldn't the mean be performed over the axis orthogonal to the axes defining the plane of the slice?

In [detector.py](https://github.com/ymahlau/fdtdx/blob/main/src/fdtdx/objects/detectors/detector.py#L318), lines 318-358:

```
        elif squeezed_ndim == 3 and self.num_time_steps_recorded == 1:
            # single step, 3d-plot. # TODO: do as mean over planes
            for k, v in squeezed_arrs.items():
                xy_slice = squeezed_arrs[k].mean(axis=0) ### should be 2?
                xz_slice = squeezed_arrs[k].mean(axis=1)
                yz_slice = squeezed_arrs[k].mean(axis=2) ### should be 0?
                fig = plot_2d_from_slices(
                    xy_slice=xy_slice,
                    xz_slice=xz_slice,
                    yz_slice=yz_slice,
                    resolutions=(
                        self._config.resolution,
                        self._config.resolution,
                        self._config.resolution,
                    ),
                    plot_dpi=self.plot_dpi,
                    plot_interpolation=self.plot_interpolation,
                )
                figs[k] = fig
        elif squeezed_ndim == 4 and self.num_time_steps_recorded > 1:
            # video with 3d-volume in each time step. plot as slices
            for k, v in squeezed_arrs.items():
                xy_slice = squeezed_arrs[k].mean(axis=1) ### should be 3?
                xz_slice = squeezed_arrs[k].mean(axis=2)
                yz_slice = squeezed_arrs[k].mean(axis=3) ### should be 1?
                path = generate_video_from_slices(
                    plt_fn=plot_from_slices,
                    xy_slice=xy_slice,
                    xz_slice=xz_slice,
                    yz_slice=yz_slice,
                    progress=progress,
                    num_worker=self.num_video_workers,
                    resolutions=(
                        self._config.resolution,
                        self._config.resolution,
                        self._config.resolution,
                    ),
                    plot_dpi=self.plot_dpi,
                    plot_interpolation=self.plot_interpolation,
                )
                figs[k] = path
```

renaissancenerd:
> Hi [@rachsmith1](https://github.com/rachsmith1), thanks for catching this! This is definitely a bug. It is a bit concerning that this was never noticed by us before.
> 
> [@renaissancenerd](https://github.com/renaissancenerd), can you make a quick PR to fix this bug?

Yes, no problem.

PR #223, opened 2026-01-13
fix for the detector plotting bug

renaissancenerd:
For #220 , each slice plane should average over the axis that is orthogonal (perpendicular) to that plane


ymahlau:
very minor thing, but can you put the entry in changelog.md in the "changed" sections, not "added". I think added should be mainly new features, not bugfixes

PR #233, opened 2026-01-18
added functionality for partial_real_position and unit tests

### LINKED ISSUE #226: fix partial_real_position
Currently, partial_real_position can be specified in an Object, but it is never used during object placement. This is clearly a bug.

In resolve_object_constraints (https://github.com/ymahlau/fdtdx/blob/d004b761a75dc5882bb240dec8cea1bf1c94e65a/src/fdtdx/fdtd/initialization.py#L492), add functionality that the specification is actually used during placement. 

Also add test cases to test the behavior.

renaissancenerd:
resolved #226

### DIFF
--- a/tests/fdtd/test_initialization.py
+++ b/tests/fdtd/test_initialization.py
@@ -1934,17 +1934,6 @@ def test_extend_to_inf_lower_bound_only_already_removed_from_extension_obj(
 ):
     """Lines 1107-1118: b0 not None, b1 None, size not None, but (o, 1) already
     absent from extension_obj because the SizeExtensionConstraint loop removed it.
-
-    When the SizeExtensionConstraint loop (lines 1098-1102) runs for axis 0 and
-    removes (obj1, 1), the per-object loop at line 1116 finds the `elif` branch
-    (b0 not None, b1 None, size not None) and evaluates `if (o, 1) in extension_obj`
-    as False — it must not crash and must not double-remove.
-
-    Setup: obj1 has known size=20 on axis 0, lower bound b0=10 (set by constraint),
-    no upper bound. A SizeExtensionConstraint on axis 0 direction "+" removes (obj1,1)
-    before the per-object loop runs. Nothing else resolves → _extend_to_inf_if_possible
-    is called. The upper bound is then derived from lower+size by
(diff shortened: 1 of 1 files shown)


src/fdtdx/fdtd/initialization.py
```diff
@@ -531,6 +531,45 @@ def resolve_object_constraints(
...
+                        half_size = size / 2
+                        lower = round(grid_center - half_size)
+                        upper = lower + size  # Ensure exact size by computing from lower
+                        slice_dict[obj_name][axis][0] = lower
+                        slice_dict[obj_name][axis][1] = upper
+                    else:
+                        # If size unknown, just store the center position somehow
+                        # We'll need to resolve this later when size becomes known
+                        # For now, we can't fully resolve the position
+                        pass
```
ymahlau: please remove the if-else construct, because the else case is not used. You can keep the comment, it is a good explanation

ymahlau:
@renaissancenerd if you take a look [here](https://app.codecov.io/gh/ymahlau/fdtdx/pull/233?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=Yannik+Mahlau#d44c18803f3b6a14b2156aa1f2920d4b-R608), you can see that the remaining lines are mostly from errors. So you should create a setup that creates these errors. Your new test cases in the last commit only test for working setups, so these errors are never reached. You should fabricate a setup that reaches the missing lines.

src/fdtdx/fdtd/initialization.py
```diff
@@ -531,6 +531,99 @@ def resolve_object_constraints(
...
+
+                        # Only set if not already set, and verify consistency if partially set
+                        if b0 is None:
+                            slice_dict[obj_name][axis][0] = lower
+                            resolved_something = True
+                        elif b0 != lower:
+                            raise Exception(
+                                f"Inconsistent position for {obj_name} axis {axis}: "
+                                f"partial_real_position implies lower bound {lower}, "
+                                f"but constraint set it to {b0}"
```
ymahlau, line 622: As another important point, you should NOT raise an Exception here, but rather add the error message to the errors dictionary

PR #235, opened 2026-01-19
Add support for fully anisotropic materials

rachsmith1:
Following the excellent work by @bruxillensis in #204 enabling diagonally anisotropic materials, this PR builds on top of that work and aims to enable support for fully anisotropic materials, i.e. materials with non-zero values in the off-diagonal elements of the material property tensors.

This is fully backwards compatible and maximally memory saving, where the following arrays sizes should be used no matter how the user specifies the material properties during initialization:
- Isotropic: (1, Nx, Ny, Nz)
- Diagonally anisotropic: (3, Nx, Ny, Nz)
- Fully anisotropic: (9, Nx, Ny, Nz)

I have added additional test cases and a new example that showcases the properties of a fully anisotropic material.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,16 +61,20 @@ export NCCL_PROTO="SIMPLE,LL,LL128"
 ## Citation
 If you find this repository helpful for you work, please consider citing:
 ```
-@article{schubert2025quantized,
-  title={Quantized inverse design for photonic integrated circuits},
-  author={Schubert, Frederik and Mahlau, Yannik and Bethmann, Konrad and Hartmann, Fabian and Caspary, Reinhard and Munderloh, Marco and Ostermann, J{\"o}rn and Rosenhahn, Bodo},
-  journal={ACS omega},
-  volume={10},
-  number={5},
-  pages={5080--5086},
-  year={2025},
-  publisher={ACS Publications}
+@article{Mahlau2026,
+  doi = {10.21105/joss.08912},
+  url = {https://doi.org/10.21105/joss.08912},
+  year = {2026},
+  publisher = {The Open Journal},
+  volume = {11},
+  number = {117},
+  pages = {8912},
+  author = {Mahlau, Yannik and Schubert, Frederik and Berg, Lukas and Rosenhahn, Bodo},
+  title = {FDTDX: High-Performance Open-Source FDTD Simulation with Automatic Differentiation},
+  journal = {Journal of Open Source Software}
 }
+
+
 ```
 
 ## Acknowedgement
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -14,6 +14,7 @@ API
     fdtdx.BrushConstraint2D
     fdtdx.circular_brush
     fdtdx.ClosestIndex
+    fdtdx.colors.Color
     fdtdx.compute_energy
     fdtdx.compute_mode
     fdtdx.compute_poynting_flux
@@ -57,6 +58,12 @@ API
     fdtdx.PillarDiscretization
     fdtdx.place_objects
     fdtdx.plot_setup
+    fdtdx.plot_setup_from_side
+    fdtdx.plot_setup_from_side
+    fdtdx.plot_material_from_side
+    fdtdx.plot_material
+    fdtdx.plot_field_slice_component
+    fdtdx.plot_field_slice
     fdtdx.PositionConstraint
     fdtdx.PoyntingFluxDetector
     fdtdx.private_field
@@ -84,4 +91,4 @@ API
     fdtdx.UniformPlaneSource
     fdtdx.WaveCharacter
     fdtdx.wavelength_to_period
-
+    
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -53,16 +53,17 @@ Citation
 If you find this repository helpful for your work, please consider citing:
 
 .. code-block:: bibtex
-
-   @article{schubert2025quantized,
-      title={Quantized inverse design for photonic integrated circuits},
-      author={Schubert, Frederik and Mahlau, Yannik and Bethmann, Konrad and Hartmann, Fabian and Caspary, Reinhard and Munderloh, Marco and Ostermann, J{\"o}rn and Rosenhahn, Bodo},
-      journal={ACS omega},
-      volume={10},
-      number={5},
-      pages={5080--5086},
-      year={2025},
-      publisher={ACS Publications}
+   @article{Mahlau2026,
+      doi = {10.21105/joss.08912},
+      url = {https://doi.org/10.21105/joss.08912},
+      year = {2026},
+      publisher = {The Open Journal},
+      volume = {11},
+      number = {117},
+      pages = {8912},
+      author = {Mahlau, Yannik and Schubert, Frederik and Berg, Lukas and Rosenhahn, Bodo},
+      title = {FDTDX: High-Performance Open-Source FDTD Simulation with Automatic Differentiation},
+      journal = {Journal of Open Source Software}
    }
 
 
--- a/src/fdtdx/__init__.py
+++ b/src/fdtdx/__init__.py
@@ -1,4 +1,5 @@
 from fdtdx import constants
+from fdtdx.colors import Color
 from fdtdx.config import GradientConfig, SimulationConfig
 from fdtdx.constants import wavelength_to_period
 from fdtdx.conversion.json import export_json, export_json_str, import_from_json
@@ -19,7 +20,6 @@ from fdtdx.core.physics.metrics import (
     normalize_by_poynting_flux,
 )
 from fdtdx.core.physics.modes import compute_mode
-from fdtdx.core.plotting import colors
 from fdtdx.core.switch import OnOffSwitch
 from fdtdx.core.wavelength import WaveCharacter
 from fdtdx.fdtd.backward import full_backward
@@ -47,7 +47,6 @@ from fdtdx.objects.device.parameters.continuous import (
     StandardToPlusOneMinusOneRange,
 )
 from fdtdx.objects.device.parameters.discrete import (
-    BOTTOM_Z_PADDING_CONFIG_REPEAT,
     BinaryMedianFilterModule,
     ConnectHolesAndStructures,
     RemoveFloatingMaterial,
@@ -86,7 +85,9 @@ from fdtdx.objects.static_material.polygon import ExtrudedPolygon
 from fdtdx.objects.static_material.sphere import Sphere
 from fdtdx.objects.static_material.static import SimulationVolume, UniformMaterialObject
 from fdtdx.utils.logger import Logger
-from fdtdx.utils.plot_setup import plot_setup
+from fdtdx.utils.plot_field_slice import plot_field_slice, plot_field_slice_component
+from fdtdx.utils.plot_material import plot_material, plot_material_from_side
+from fdtdx.utils.plot_setup import plot_setup, plot_setup_from_side
 
 __all__ = [
     "Detector",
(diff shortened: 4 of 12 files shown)


src/fdtdx/core/physics/curl.py
```diff
@@ -133,76 +133,20 @@ def curl_E(
...
 
         # Update auxiliary fields
-        psi_Hxy = b_y * psi_Hxy + a_y * dyEz
-        psi_Hxz = b_z * psi_Hxz + a_z * dzEy
-        psi_Hyz = b_z * psi_Hyz + a_z * dzEx
-        psi_Hyx = b_x * psi_Hyx + a_x * dxEz
-        psi_Hzx = b_x * psi_Hzx + a_x * dxEy
-        psi_Hzy = b_y * psi_Hzy + a_y * dyEx
+        psi_Hxy = b[4, :, :, :] * psi_Hxy + a[4, :, :, :] * dyEz
+        psi_Hxz = b[5, :, :, :] * psi_Hxz + a[5, :, :, :] * dzEy
```
ymahlau: minor comment, you could remove the colons ":, :, :" here for readability

src/fdtdx/core/physics/modes.py
```diff
@@ -125,14 +126,24 @@ def compute_mode(
...
     if isinstance(inv_permeabilities, jax.Array) and inv_permeabilities.ndim > 0:
         if (
             not (
-                inv_permeabilities.ndim == 4 and (inv_permeabilities.shape[0] == 1 or inv_permeabilities.shape[0] == 3)
+                inv_permeabilities.ndim == 4
+                and (
+                    inv_permeabilities.shape[0] == 1
+                    or inv_permeabilities.shape[0] == 3
+                    or inv_permeabilities.shape[0] == 9
+                )
```
ymahlau: again minor comment, but could you try to format this a bit better for readability? For example extract the condition to a separate line beforehand

src/fdtdx/core/grid.py
```diff
@@ -100,20 +131,54 @@ def _raise_if_anisotropic(is_iso):
...
+                    ),
+                    jnp.logical_and(
+                        jnp.logical_and(
+                            jnp.allclose(inv_permeabilities[6], 0.0), jnp.allclose(inv_permeabilities[7], 0.0)
+                        ),
+                        inv_permeabilities.shape[0] == 9,
+                    ),
+                )
+
+                is_isotropic = jnp.logical_or(is_diagonal_isotropic, is_full_isotropic)
```
ymahlau: I think at this point it might be helpful to separate into helper functions somewhere in core/physics. You can write this a bit more beautiful using bitwise and ("&") and bitwise or ("|"). Maybe something like `def is_diagonal_anisotropic(permittivity)` and `def is_full_anisotropic(...`

src/fdtdx/core/physics/modes.py
rachsmith1: Fyi some of the formatting was done by the pre-commit. I can change them for readability but they might fail the pre-commit.

src/fdtdx/fdtd/initialization.py
```diff
@@ -397,7 +516,11 @@ def _init_arrays(
 
             if magnetic_conductivity is not None:
                 allowed_conds = jnp.asarray(
-                    compute_allowed_magnetic_conductivities(o.materials, isotropic=isotropic_magnetic_conductivity)
+                    compute_allowed_magnetic_conductivities(
+                        o.materials,
+                        isotropic=isotropic_magnetic_conductivity,
+                        diagonally_anisotropic=diagonally_anisotropic_magnetic_conductivity,
+                    )
```
ymahlau, line 523: This is not a cause for this PR, but the length of this function is getting a bit out of hand. After merging the PR, we will need to refactor.

src/fdtdx/fdtd/update.py
```diff
@@ -144,21 +309,188 @@ def _update():
...
+            B[0, 0] * curl[0] + B[0, 1] * curly_x_avg + B[0, 2] * curlz_x_avg
+        )
+        Ey = (A[1, 0] * Ex_y_avg + A[1, 1] * E[1] + A[1, 2] * Ez_y_avg) - (
+            B[1, 0] * curlx_y_avg + B[1, 1] * curl[1] + B[1, 2] * curlz_y_avg
+        )
+        Ez = (A[2, 0] * Ex_z_avg + A[2, 1] * Ey_z_avg + A[2, 2] * E[2]) - (
+            B[2, 0] * curlx_z_avg + B[2, 1] * curly_z_avg + B[2, 2] * curl[2]
+        )
+
+        E = jnp.stack((Ex, Ey, Ez), axis=0)
```
ymahlau, line 280: Can you explain the code here? I am no expert on the fully anisotropic materials, so for me it is very difficult to review this part.

Another comment: I would like to keep the update functions as simple as possible to make further development and maintainance as easy as possible. Could you try to refactor this very lengthy code into separate functions?

src/fdtdx/fdtd/update.py
```diff
@@ -204,28 +537,185 @@ def update_H(
...
+            B[0, 0] * curl[0] + B[0, 1] * curly_x_avg + B[0, 2] * curlz_x_avg
+        )
+        Hy = (A[1, 0] * Hx_y_avg + A[1, 1] * arrays.H[1] + A[1, 2] * Hz_y_avg) - (
+            B[1, 0] * curlx_y_avg + B[1, 1] * curl[1] + B[1, 2] * curlz_y_avg
+        )
+        Hz = (A[2, 0] * Hx_z_avg + A[2, 1] * Hy_z_avg + A[2, 2] * arrays.H[2]) - (
+            B[2, 0] * curlx_z_avg + B[2, 1] * curly_z_avg + B[2, 2] * curl[2]
+        )
+
+        H = jnp.stack((Hx, Hy, Hz), axis=0)
```
ymahlau, line 397: same here

src/fdtdx/materials.py
```diff
@@ -37,39 +69,55 @@ class Material(TreeClass):
...
-    #: Stored internally as a 3-tuple. Defaults to (1.0, 1.0, 1.0).
-    permittivity: tuple[float, float, float] = frozen_field(
-        default=(1.0, 1.0, 1.0),
+    #: For isotropic materials, provide a scalar float.
+    #: For diagonally anisotropic materials, provide a tuple of 3 floats (εx, εy, εz).
+    #: For fully anisotropic materials, provide either:
+    #:   - A tuple of 9 floats (εxx, εxy, εxz, εyx, εyy, εyz, εzx, εzy, εzz), or
+    #:   - A nested tuple ((εxx, εxy, εxz), (εyx, εyy, εyz), (εzx, εzy, εzz))
+    #: Stored internally as a 9-tuple. Defaults to (1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0).
+    permittivity: tuple[float, float, float, float, float, float, float, float, float] = frozen_field(
```
ymahlau, line 77: Do i see correctly, that users can still specify a single float or a tuple of three floats here? In that case the typing should reflect that: float | tuple[float, float, float] | tuple[float, ...

src/fdtdx/materials.py
```diff
@@ -93,13 +141,145 @@ def __init__(
...
+    @overload
     def __init__(
         self,
         *,
-        permittivity: float | tuple[float, float, float] = (1.0, 1.0, 1.0),
-        permeability: float | tuple[float, float, float] = (1.0, 1.0, 1.0),
-        electric_conductivity: float | tuple[float, float, float] = (0.0, 0.0, 0.0),
-        magnetic_conductivity: float | tuple[float, float, float] = (0.0, 0.0, 0.0),
+        permittivity: tuple[float, float, float, float, float, float, float, float, float] = (
+            1.0,
```
ymahlau: what is the purpose of overriding the init function here? I think this could lead to problems.

src/fdtdx/core/physics/modes.py
```diff
@@ -125,14 +126,24 @@ def compute_mode(
     """
     # Input validation
     if (
-        not (inv_permittivities.ndim == 4 and (inv_permittivities.shape[0] == 1 or inv_permittivities.shape[0] == 3))
+        not (
+            inv_permittivities.ndim == 4
+            and (
+                inv_permittivities.shape[0] == 1 or inv_permittivities.shape[0] == 3 or inv_permittivities.shape[0] == 9
+            )
```
z-y-n:
```suggestion
            and inv_permittivities.shape[0] in (1, 3, 9)
```
@rachsmith1 something like this would improve readability a bit, or just move it to some bool variables.  and likewise for permeabilities too

src/fdtdx/core/physics/curl.py
rachsmith1: Fixed in latest commit

src/fdtdx/core/physics/modes.py
rachsmith1: Formatted better for readability in latest commit

src/fdtdx/core/physics/modes.py
rachsmith1: Formatted better for readability in latest commit

src/fdtdx/materials.py
rachsmith1: Fixed in latest commit

src/fdtdx/core/grid.py
rachsmith1: Better formatting in latest commit

src/fdtdx/fdtd/update.py
rachsmith1: Fixed in latest commit

src/fdtdx/materials.py
rachsmith1: Fixed in latest commit. Yes, the user can specify a float, a 3-tuple, a 9-tuple, or a 3-tuple of 3-tuples, and they are normalized by _normalize_material_property and stored as a 9-tuple. It should be completely backwards compatible and there are plenty of test cases in test_anisotropic_materials.py.

src/fdtdx/fdtd/update.py
rachsmith1: The fully anisotropic material tensors have non-zero off-diagonal elements, so this means that any single component of the E and H fields (e.g. Ex) depends on all three components in the update (e.g. all Ex, Ey, and Ez) so you can't just do a simple element-wise calculation. However, these vector components are staggered on the Yee grid and do not exist at the same point in space. In order to get these values at the exact same location for the update, the easiest thing to do is interpolate using a 4-point spatial average using adjacent Yee points. 

I have added helper functions and comments, hopefully it is easier to follow.

rachsmith1:
@ymahlau Is there anything else you want to discuss re this PR?

ymahlau:
Hi @rachsmith1, sorry that I am so inactive here. I will take a closer look at this mid-next week and give a proper review

src/fdtdx/fdtd/misc.py
```diff
@@ -68,3 +69,120 @@ def add_boundary_interfaces(
...
+    """
+
+    return (
+        (
+            field[component]
+            + jnp.roll(field[component], 1, axis=location)
+            + jnp.roll(field[component], -1, axis=component)
+            + jnp.roll(field[component], (1, -1), axis=(location, component))
+        )
+        / 4
```
ymahlau, line 187: Minor comment: could you please move these functions into a new file in fdtdx/core/physics/anisotopric.py

The misc file in fdtd only exists because it has a dependency on the arraycontainer, which would create circular dependencies if we put in in the core module. Your new functions are more similar to the curl function, where we only work with the pure jax arrays

PR #241, opened 2026-02-09
Add option to export to VTI file for visualization with external tools

l-berg:
This PR implements the option to export jax or numpy arrays in the [VTI](https://docs.vtk.org/en/latest/vtk_file_formats/vtkxml_file_format.html) (VTK ImageData) format. This format is specialized for equi-distant grids and supports vector data such as E- and H-fields. This makes simulation data accessible to external tools (including Paraview, VisIt and Mayavi) for further analysis and debugging. Here is an example:

image

This plot was generated with ParaView. The data was generated using the [mode example notebook](https://github.com/ymahlau/fdtdx-notebooks/blob/main/advanced/modes.ipynb) and exported to three files using this snippet:
```python
from pathlib import Path
from fdtdx import export_vti, export_arrays_snapshot_to_vti

path = Path('/tmp')
export_arrays_snapshot_to_vti(arrays, path / 'simulation_snapshot.vti', config.resolution)

source = new_objects.sources[0]
export_vti({'E_source': source._E}, path / 'mode_source.vti', config.resolution, grid_slice=source.grid_slice)

detector = new_objects.detectors[1]
export_vti({'E_detector': jnp.real(detector._mode_E)}, path / 'mode_detector.vti', config.resolution, grid_slice=detector.grid_slice)
```

This PR implements the VTI format from scratch without any external dependencies. Of course, the `vtk` package (and packages that build on it, like `pyvista`) would significantly simplify the export process, but they are >100 MB dependencies, and I did not find smaller projects. The popular `pyevtk` package does not seem active and has license issues.

In the future, it would be straightforward to support recti-linear grids by adding the VTR format which simply holds three additional 1d arrays for the the x, y, and z spacing.

The export interface exposes export_vti (mimicking export_stl) and the convenience function export_arrays_snapshot_to_vti (alternative name suggestions welcome).

### DIFF
--- a/tests/conversion/test_vti.py
+++ b/tests/conversion/test_vti.py
@@ -1,13 +1,13 @@
-import pytest
-
 import jax
 import jax.numpy as jnp
+import pytest
 
 import fdtdx
 from fdtdx import (
-    export_vti, 
-    export_arrays_snapshot_to_vti, 
+    export_arrays_snapshot_to_vti,
+    export_vti,
 )
+
 
 def test_export_vti_file_creation(tmp_path):
     """Test that export_vti creates a valid file with expected XML tags."""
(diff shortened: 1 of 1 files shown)


ymahlau:
@l-berg Also please add the new functions to the docs -> api list in the alphabetically correct position

Issue #246, opened 2026-02-18
Add Simulation from json function


ymahlau:
The output of the fdtdx GUI (https://github.com/ymahlau/fdtdx-studio) is a JSON containing the specification of config, SimulationVolume, list of constraints and object list. It should be possible to run a simulation without writing a python script using fdtdx.

Please add the following function:
A function that takes a path to a json file as input, then constructs the corresponding python fdtdx SImulationObjects, constraints etc. 
Then the function runs a simulation using these objects and saves the results. 

Please add for every component (loading, python object building, simulation, saving) a separate function that can also be called individually for better re-usability.
Also write test cases to check functionality.

PR #251, opened 2026-02-26
fix import json

sIHURs:
tackles #246
Fixed a bug where importing JSON couldn’t properly handle dataclass types, and added unit tests for JSON export/import as well as running a simulation from imported JSON.

---
## Summary by cubic
Fixes JSON export/import to correctly handle dataclasses, dicts, and sequences, enabling full simulations to run from imported JSON. Addresses #246.

- **Bug Fixes**
  - Export dataclasses with type info and a field dict; skip private attrs.
  - Reconstruct dataclasses via cls(**kwargs) when __value__ is a dict.
  - Correct dict handling (use name == "dict").
  - Import sequences without dropping items.

- **Tests**
  - Add round-trip JSON tests for SimulationConfig, SimulationVolume, sources, objects, detectors, constraints (with reference integrity), and ObjectContainer.
  - Add end-to-end test that exports settings, imports them, places objects, JIT-compiles, and runs an FDTD simulation successfully.

<sup>Written for commit b2758c9b3e90638c5d11ee2520e42c0bffcb3aef. Summary will update on new commits.</sup>

### DIFF
--- a/src/fdtdx/conversion/json.py
+++ b/src/fdtdx/conversion/json.py
@@ -114,6 +114,7 @@ def _import_obj_from_json(obj: dict | float | int | str | bool | None) -> Any:
         if isinstance(vals, dict):
             kwargs = {k: _import_obj_from_json(v) for k, v in vals.items()}
             return cls(**kwargs)
+        # sequence
         imported_vals = [_import_obj_from_json(v) for v in vals]
         return cls(imported_vals)
     # dictionary


src/fdtdx/conversion/json.py
```diff
@@ -104,10 +111,14 @@ def _import_obj_from_json(obj: dict | float | int | str | bool | None) -> Any:
     # sequence
```
l-berg: This comment now outdated, since sequences AND dataclasses are handled here.

sIHURs: I've removed the outdated comment

PR #254, opened 2026-03-01
fix for bug in plot_material

### LINKED ISSUE #248: plot_material does not show structures
In some situations, the plot_material function does not seem to correctly display the materials in the simulation. This needs investigation and fixing.

renaissancenerd:
resolves #248 

---
## Summary by cubic
Fixes incorrect slicing in plot_material that collapsed plot extents and produced wrong slices. Adds a material_axis parameter to select x/y/z components and forwards it to all subplots; resolves #248.

- **Bug Fixes**
  - Slice out the component dimension first (material_array[material_axis] → (Nx, Ny, Nz)).
  - Compute extents from spatial dimensions (spatial_shape) instead of component count.
  - Correct per-view indexing: [:, :, k], [:, j, :], [i, :, :].
  - Tests verify full-domain extents across all views and material_axis selections.

- **New Features**
  - plot_material accepts material_axis (0, 1, 2) and passes it through; default is 0.

<sup>Written for commit d840b0607e08b635a8e5bb5eeab5a0bbbd3459e9. Summary will update on new commits.</sup>

### DIFF
--- a/src/fdtdx/utils/plot_material.py
+++ b/src/fdtdx/utils/plot_material.py
@@ -31,7 +31,7 @@ def plot_material_from_side(
         config (SimulationConfig): Configuration object containing simulation parameters like resolution
         arrays (ArrayContainer): Container holding the material arrays (permittivity, permeability)
         viewing_side (Literal['x', 'y', 'z']): Which plane to view ('x' for YZ, 'y' for XZ, 'z' for XY)
-        materialaxis (int): Which material axis to plot (for anisotropic materials). Can be 0, 1, 2 for x, y or z.
+        material_axis (int): Index into the leading component dimension of the material array (for anisotropic materials).
         filename (str | Path | None, optional): If provided, saves the plot to this file instead of displaying
         ax (Any | None, optional): Optional matplotlib axis to plot on. If None, creates new figure
         plot_legend (bool, optional): Whether to add a colorbar legend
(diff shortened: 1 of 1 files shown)


renaissancenerd:
test_plot_material_all_permittivity
test_plot_material_all_types
test_plot_material_cylinder
test_plot_material_sphere

src/fdtdx/utils/plot_material.py
```diff
@@ -67,16 +67,20 @@ def plot_material_from_side(
     # Calculate slice index from position
     slice_offset = round(position / config.resolution)
 
-    # Get array shape
-    array_shape = material_array.shape
+    # Select the requested component axis, giving a 3D array of shape (Nx, Ny, Nz).
```
l-berg: I think such a long comment should not be part of the main source files. And changing the dimensions of `material_array` might be confusing. How about this approach: Rename `array_shape` to `spatial_shape`, which is simply `spatial_shape = material_array.shape[1:]`.

src/fdtdx/utils/plot_material.py
l-berg: While at it, maybe you could also fix this typo: `materialaxis` should be `material_axis`.

l-berg: Additionally, the `0, 1, 2 for x, y or z` is outdated since it only covers diagonal and not fully tensorial anisotropic materials.

PR #255, opened 2026-03-01
Enable running fdtdx simulations directly from GUI JSON export

sIHURs:
tackles #246

Based on the [fdtdx example](https://github.com/sIHURs/fdtdx-notebooks/blob/main/examples/simulation_example.ipynb), users can now import or export JSON to run and execute the simulation. It can also later accept the JSON output generated by fdtdx-studio.

```python
settings = JsonSetup.load_json(json_file_path)
seed = settings.meta.get("seed")
key, subkey = jax.random.split(key)

objects, arrays, params, cfg2, _ = place_objects(
        object_list=setup2.object_list,
        config=setup2.config,
        constraints=setup2.constraints,
        key=subkey,
    )

def sim_fn(params: ParameterContainer, arrays: ArrayContainer, key: jax.Array):
        # Apply parameters to objects and arrays
        arrays, new_objects, _ = apply_params(arrays, objects, params, key)

        # Run FDTD simulation (forward)
        final_state = run_fdtd(arrays=arrays, objects=new_objects, config=cfg2, key=key)
        _, arrays = final_state
        return arrays
compiled = jax.jit(sim_fn).lower(params, arrays, key).compile()
out = compiled(params, arrays, subkey)

    ...

```

---
## Summary by cubic
Run FDTD simulations directly from a GUI-exported JSON via `JsonSetup` and an extended JSON codec. Load config, objects, and constraints from JSON and run end-to-end. Addresses #246.

- **New Features**
  - Added `JsonSetup` with load_json/loads/from_dict/dumps/export_json, optional `meta` (e.g., seed), and validation (list types, exactly one SimulationVolume, unique names, valid object/constraint types, constraint references).
  - JSON dumps include schema `fdtdx.place_objects.v1` and version `0.6.0`; codec now serializes dataclasses and unwraps list-wrapper formats; file-based export/import supported.
  - Tests cover serializing config, volume, sources, objects, constraints, and running a jitted simulation from exported JSON and a `JsonSetup` file.

- **Bug Fixes**
  - Fixed dict import check (use name == "dict") and made sequence decoding consistent.
  - Properly import dataclasses carried in `__value__` and handle list/dict wrapper edge cases.

<sup>Written for commit 470a296486d4590ab79e55f62329b0ba3cdc7883. Summary will update on new commits.</sup>

### DIFF
--- a/src/fdtdx/conversion/json.py
+++ b/src/fdtdx/conversion/json.py
@@ -60,8 +60,6 @@ class JsonSetup:
             raise ValueError("Missing 'object_list'.")
 
         constraints = data.get("constraints", [])
-        if constraints is None:
-            raise ValueError("Missing 'constraints'.")
 
         meta = data.get("meta", None)
 


src/fdtdx/conversion/json.py
```diff
@@ -1,14 +1,195 @@
...
+        config = data.get("config", data.get("SimulationConfig", None))
+        if config is None:
+            raise ValueError("Missing 'config' or 'SimulationConfig'.")
+
+        object_list = data.get("object_list", None)
+        if object_list is None:
+            raise ValueError("Missing 'object_list'.")
+
+        constraints = data.get("constraints", [])
+        if constraints is None:
```
l-berg: This will never be True, right (because constraints defaults to `[]`)? In general, the pattern of `.get(default=None)` and then checking for None afterwards seems fragile to me.

l-berg: Or is it necessary because the config may contain `None` values?

src/fdtdx/conversion/json.py
```diff
@@ -1,14 +1,195 @@
...
+
+    def dumps(self) -> str:
+        """
+        Serialize setup to JSON string.
+
+        Formatting depends on `export_json_str` implementation.
+        """
+        export_format = {
+            "__schema__": "fdtdx.place_objects.v1",
+            "__fdtdx_version__": "0.6.0",
```
l-berg, line 81: I agree that a whitelist for the objects is fine for now. But we need definitely need a better way to specify the version in the future (instead of hard-coding this string).

src/fdtdx/conversion/json.py
sIHURs: resolved in new commit, constraints can be an empty list, so the check isn't needed my mistake. I just wanted to add a simple validation to ensure config and object_list exist, which is required for the simulation to run.  The JSON structure is  fixed and is only exported by fdtdx or the GUI, so it should generally always be valid unless someone manually edits it. But the current JSON format isn't very readable acutally, we normally wouldn’t need to modify it anyway, I think

Issue #127, opened 2025-09-01
Add a dipole source to the library


georgebu31:
Today's tasks in photonics require the use of dipole sources to simulate emitters occurring in materials, 2D materials, and so on.
It would be very convenient to add a class of dipole sources to the library for later use in simulations and optimizations.

ymahlau:
Hi @georgebu31, could you explain how the results are difficult to interpret or diverging?

I think the easiest way to share the code would be to open a draft PR. Alternatively, you can also share by mail (mahlau(at)tnt.uni-hannover.de).

Since I am currently doing an internship, I do not have much spare time. I will be happy to take a look at your code to see if I can spot anything immediate, but will not be able to do a deep dive. If I am unable to spot anything I can connect you to a colleague who can take a deeper look.

PR #270, opened 2026-04-06
Implement dipole source, add simulation source tests

bruxillensis:
Not too complicated of a PR this time: Implemented electric and magnetic dipoles, added simulation tests for all sources.

---
## Summary by cubic
Added a soft point dipole source (electric or magnetic) with arbitrary 3D orientation, plus comprehensive physics tests for all sources. Also added internal JAX/fdtdx skills docs under `.claude/skills` (now include detector guidance) and streamlined `CLAUDE.md`.

- **New Features**
  - Implemented `PointDipoleSource` with soft E/H injection; orientation via `polarization` + `azimuth_angle`/`elevation_angle`; supports `amplitude` and temporal profiles; exported in `fdtdx.__init__`.

- **New Tests**
  - Dipoles: free-space symmetry and radiation pattern, dielectric power scaling (P ≈ n), intermediate-field decay, E–H duality, and rotated orientations.
  - Plane sources: `GaussianPlaneSource` spatial profile and total power; `UniformPlaneSource` amplitude scaling and spatial uniformity.
  - Mode source: `ModePlaneSource` propagation (n_eff) and confinement.
  - Temporal profiles: CW steady-state phasor and Gaussian pulse bandwidth.
  - Unit tests: initialization, angle handling, E/H updates, amplitude scaling, inverse behavior.

<sup>Written for commit 8d709cb825eeeb71bb04db614c5d55c4b048bb41. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a configurable point-dipole source for electromagnetic simulations (electric/magnetic modes, polarization/orientation, amplitude control).

* **Documentation**
  * Consolidated and expanded framework guidance into dedicated skill-style docs and replaced a dense architecture section with pointers to the new guide.

* **Tests**
  * Large suite of new unit and end-to-end physics tests validating dipole, plane, Gaussian, mode, and temporal-source behaviors and expected radiation/field metrics.


bruxillensis:
I realized I did not do arbitrary rotations of the dipole. Might be good to include in this PR.

bruxillensis:
Example output (log scale), just because I think it looks fun :)

https://github.com/user-attachments/assets/5fb466b4-1154-459b-b7bd-15e7ec426ecb

ymahlau:
Awesome! Will give a review later today :)

ymahlau:
@cubic-dev-ai review

src/fdtdx/objects/sources/dipole.py
```diff
@@ -0,0 +1,155 @@
...
+            (0=x, 1=y, 2=z) before angle rotations are applied.
+        azimuth_angle: Rotation around the vertical axis in degrees
+            (default 0).
+        elevation_angle: Rotation around the horizontal axis in degrees
+            (default 0).
+        source_type: ``"electric"`` injects current into the E update,
+            ``"magnetic"`` injects into the H update.
+        amplitude: Source strength (default 1.0). Combined with
+            ``static_amplitude_factor`` and the temporal profile to compute the
+            injected current at each time step.
```
ymahlau: Minor comment: Usually we keep the argument definition directly at the argument definition (as you have done below), so there is no need to have an args section in the docstring. In the sphinx-rendered docs this would lead to a duplicated section (where for the args-section the parameters of the superclass are missing).

tests/simulation/physics/sources/test_plane_source.py
```diff
@@ -0,0 +1,119 @@
...
+    arrays_2x = _build_and_run(amplitude_factor=2.0)
+
+    # Extract Ex phasor (may have trailing dims even with reduce_volume)
+    p1 = complex(arrays_1x.detector_states["phasor"]["phasor"][0, 0].ravel()[0])
+    p2 = complex(arrays_2x.detector_states["phasor"]["phasor"][0, 0].ravel()[0])
+
+    assert abs(p1) > 1e-30, "1× phasor amplitude is zero"
+
+    ratio = abs(p2) / abs(p1)
+    assert abs(ratio - 2.0) / 2.0 < 0.05, f"Amplitude ratio: {ratio:.3f}, expected 2.0"
```
ymahlau, line 120: This might be a bit of of nitpicking in a really good PR: I think there may be more useful tests than just testing the doubling of amplitude. For example, output amplitude and phase should be the same everywhere in xy-direction. But feel free to ignore this comment if you dont feel like changing it, this also good as is.

tests/simulation/physics/sources/test_gaussian_source.py
```diff
@@ -0,0 +1,273 @@
...
+        dtype=jnp.float32,
+    )
+    objects, constraints = [], []
+
+    volume = fdtdx.SimulationVolume(
+        partial_real_shape=(domain_xy, domain_xy, _DOMAIN_Z),
+    )
+    objects.append(volume)
+
+    bound_cfg = fdtdx.BoundaryConfig.from_uniform_bound(
```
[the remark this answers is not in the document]
ymahlau: dont feel like this is a valid criticism

ymahlau:
@bruxillensis this looks very good to me! Only some minor comments above.

bruxillensis:
Sounds good! Agreed with those comments. Also, do you mind if I push some Claude skills with this pr? I added a few and they're relatively helpful.

ymahlau:
Not at all, put them in the PR!

bruxillensis:
Pushed! While I work on the comments, here is what I added:

**3 knowledge skills**
jax
jax-pde
fdtdx

These are knowledge-only skills that are loaded automatically when launching claude. They help with the structure/context of fdtdx, jax, and doing physics with jax.

**1 user involkable skill**
Putting /implement before a prompt will cause claude to use the implement skill. It does an implementation in 4 phases:
1. Drafts a preliminary plan based on its knowledge of fdtdx (knowledge skill)
2. Enters planning mode and spawns some subagents to gather info on the codebase based on prelim plan
3. Finalizes the implementation plan
4. Clears context and implements (using parallel subagents if there is need)

You can use it for more complicated requests. jax and jax-pde were taken from https://github.com/tondevrel/scientific-agent-skills.

Oh, and I modified claude.md to put fdtdx library information in the fdtdx skill. Claude.md is still used for instructions on initializing, testing, style, etc.

tests/simulation/physics/sources/test_plane_source.py
bruxillensis: Good comment! I added a uniformity test

ymahlau:
also relates to #127

ymahlau:
@bruxillensis thanks for including the skills! I have a couple of questions/comments:
- Do the jax and jax-pde skills actually increase performance? I feel like the pde information does not really apply to fdtdx and the jax information is very basic information that the models should already know on their own. I would guess that these skills only waste tokens, but if you made a different experience let me know. In my opinion we should not force these skills on a user by including them directly in fdtdx.
- In addition to the points above, I think you will need to remove the skills because they have an Apache2.0 license, which we cannot include in the MIT-licensed fdtdx. 
- Independently from the skills, your last PR introduced a lot of duplicated code in the simulation test (I agree with AI-review here). Could you please clean this up a bit before we accept the PR?

The fdtdx and knowledge skill as well as the plan-and-implement look very useful! Thanks for sharing them!

bruxillensis:
Re skills: My bad! I didnt realize Apache would conflict. I also dont see a quantititative difference with the jax skills vs what claude already knows in general. I'll remove them. Ill fix up the test as well.

bruxillensis:
@ymahlau I think this is ready to be merged. The two comments left don't seem like big issues to me, but let me know if you think they should be addressed. Note: pre-commit is failing because of a type error in mode.py, but I think its from a different PR.

ymahlau:
Thanks @bruxillensis, this PR looks ready to merge to me as well!. The failing checks come (probably?) from an update of the typechecker, because we did not touch that part of the code in a while. Definitely unrelated to this PR.

PR #294, opened 2026-04-18
Dispersive Material Implementation

bruxillensis:
*Draft* implementation #4

  - Adds a new fdtdx.dispersion module with pole classes (Drude, Lorentz, etc.), ADE state plumbed through fdtd/{container,initialization,forward,update}.py, and material-side integration in materials.py. Should be extendable to other types of poles that can use the ADE equation.
  - Extends automatic differentiation to strongly dispersive regimes: reverse-mode works with dispersive ε(ω), validated by new time-reversal and gradient tests in tests/simulation/fdtd/test_time_reversal.py.
  - Convolution-based source scaling for broadband simulations in linear_polarization, mode, and tfsf sources, with unit + simulation tests.
  - Dispersion is gated, its not included if poles are not added to a material.
  - Adding dispersion to the gradient is also gated, its only needed in the strongly dispersive regime when doing inverse design. Explicit flag in the gradient config is added.
  - New example: examples/dispersive_gaussian_pulse.py. Simulates a gaussian pulse in silicon using a lossless Lorentz pole.
  - Bug fixes: I noticed the reverse update of E/H with conductive materials was incorrect. We didnt have any tests for the reversability of this. Also updated dipole 9-element tensor handling and fixed the Planning mode on the implement skill.

What this doesn't cover:
  - Dispersive anisotropic materials.
  - Complex material properites (currently not well handled, even for throwing errors).

Despite my efforts, its still about 50% slower. I know this is expected but if anyone has feedback, it's welcome.

dispersive_gaussian_pulse

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for dispersive materials (Lorentz/Drude poles), carrier-frequency-corrected permittivity, and impedance-corrected broadband source injection
  * Sources and detectors now handle dispersive corrections and oriented tensor coupling
  * New runnable dispersive Gaussian pulse example

* **Documentation**
  * Expanded dispersive material, ADE ordering, source/TFSF, detector, and gradient strategy guidance

* **Tests**
  * Extensive unit, integration, and physics tests for dispersion, sources, reversibility, and gradients

[Review Change Stack](https://app.coderabbit.ai/change-stack/ymahlau/fdtdx/pull/294)


ymahlau:
wow @bruxillensis, you work fast! 

It is also very satisfying to see that the time-reversibility implemented with dispersive materials. My colleagues and I were actually thinking about writing a small paper on this for different dispersive models, since we never saw this reported in literature anywhere. Since you already implemented this, would you like to cooperate on a paper about this? If you don't feel like you want to put a lot of time into a scientific paper, we could also just add you as a co-author to value your contribution. Let me know about your thoughts on this!

Independently from that, I will take a look at the PR in the next few days. It might take a bit longer, because I currently have a lot to do at work.

ymahlau:
Regarding speed, I think 50% slower is very reasonable since there is much more computation going on, as well as more memory read/write which also takes a lot of time. I did not look at the code yet, but do you mean with "dispersion is gated" in your PR description that speed does not change for simulations without dispersive model? That would be very important.

bruxillensis:
@ymahlau thanks! When I said "gated" I mean that the dispersive equation is not solved, and the arrays are not extended, without adding a pole to a material. So it should behave exactly as the nondispersive case. If you run the dispersive example you can see the speed difference. Same with adding the ade to the gradient. Unless you set differentiate_dispersion, the terms won't be added to the gradient so it's just the eps_nondispersive. This is fine and speeds things up so long as you're not simulating in a strongly dispersive regime of the material.

All that being said! I'd appreciate a close look by anyone who has time. Re paper: that sounds great! I'll message you separately.

src/fdtdx/fdtd/container.py
```diff
@@ -323,10 +383,13 @@ def reset_array_container(
     Returns:
         ArrayContainer: A new ArrayContainer with reset fields and optionally reset states.
     """
-    E = arrays.E * 0
-    arrays = arrays.aset("E", E)
-    H = arrays.H * 0
-    arrays = arrays.aset("H", H)
+    arrays = arrays.aset("E", jnp.zeros_like(arrays.E))
```
ymahlau: This code here will break the multi-GPU functionality: If the E array is larger than the memory on a single GPU it needs to be initalized in a distributed way. This is handled in the place_objects function. A multiplication with 0 keeps the distributed structure, while jnp.zeros_like does not (at least when i tested this a year ago or so...). So there is a reason we are doing the multiplication here

src/fdtdx/fdtd/fdtd.py
```diff
@@ -99,90 +189,6 @@ def reversible_fdtd_base(
         )
         return (state[0], state[1])
 
-    @jax.custom_vjp
-    def reversible_fdtd_primal(
```
ymahlau, line 106: I strongly believe that it would not be necessary to refactor the whole file structure like this. Can you try to keep the file structure as before and just add the dispersive fields where necessary? 

The differentiate_dispersion flag is a nice idea, but the way it is implemented here seems a bit convoluted. It should be possible to do this by just doing if statements within the primal, fwd and bwd fucntion. You could also leave it out as a first implementation, because most applications need differentiation through the  dispersive material anyways.

src/fdtdx/fdtd/forward.py
```diff
@@ -18,14 +18,21 @@ def forward_single_args_wrapper(
...
+    dispersive_c2: jax.Array | None = None,
+    dispersive_c3: jax.Array | None = None,
+    *,
     config: SimulationConfig,
     objects: ObjectContainer,
     key: jax.Array,
     record_detectors: bool,
     record_boundaries: bool,
     simulate_boundaries: bool,
+    arrays_template: ArrayContainer | None = None,
```
ymahlau: Could you elaborate a little why the arrays_template is necessary here? Is it necessary to include this as an argument here? Or is it just a result from the refactor of fdtd above? I would propose trying to remove it, because that would probably blow up memory quite a bit.

src/fdtdx/fdtd/update.py
```diff
@@ -164,7 +164,28 @@ def update_E(
         # standard update formula using lossless material
         # Component-wise multiplication for diagonally anisotropic materials:
         # E[i, x, y, z] = factor * E[i, x, y, z] + c * curl[i, x, y, z] * inv_eps[i, x, y, z]
-        E = factor * arrays.E + c * curl * inv_eps
+        E_prev = arrays.E
+        E = factor * E_prev + c * curl * inv_eps
```
ymahlau: It would be nice to use arrays.E oder E_prev consistently throughout the function

src/fdtdx/objects/detectors/mode.py
```diff
@@ -77,8 +77,16 @@ def apply(
         key: jax.Array,
         inv_permittivities: jax.Array,
         inv_permeabilities: jax.Array | float,
+        *,
```
ymahlau: On a general note, I would prefer not adding stars to the functions and let users have the freedom to decide if they want to use keyword arguments or not

src/fdtdx/objects/sources/linear_polarization.py
```diff
@@ -28,13 +30,41 @@ def apply(
         key: jax.Array,
         inv_permittivities: jax.Array,
         inv_permeabilities: jax.Array | float,
+        *,
```
ymahlau: same here

ymahlau:
Overall, this already looks good! Just a bit of refactoring in the places noted above would be necessary. I will start a general refactoring after the PR is merged to clean up the general structure even more. In some places the code has gotten a bit lenghty (independent of this PR) and could use some refactoring.

bruxillensis:
Hey @ymahlau I think this is good for another quick review. Thanks for catching the GPU issue - I still have a ways to go learning about jax. For the rest - I dropped differentiate_dispersion, it was just an attempt to improve speed, but it did create an odd refactoring. One note - I think there was a bug, conductivities were not passed to forward_single_args_wrapper, meaning on the backward pass the wrapper passes None to those args, making the reverse calc lossless. It increases the error somewhat in test_time_reversal, but not enough to trigger a fail. I think I will try to tighten the tolerances/refine the simulation tests.

ymahlau:
Awesome @bruxillensis! The reversible fdtd file looks good to me now! And good catch on the missing conductivities. Tell me when you updated the simulation tests, I think we can merge the PR then

bruxillensis:
@ymahlau I think this is ready to merge

ymahlau:
Awesome!

PR #304, opened 2026-04-27
Error check when objects are placed outside of simulation volume

### LINKED ISSUE #259: Error check when objects are placed outside of simulation volume
We should add a check if any objects (or part of objects) are placed outside of the simulation volume. If so, an exception should be raised in place_objects (and the error str should be set in the resolve_constraints function).

We also need test cases that verify that these errors are handled correctly.

renaissancenerd:
resolves #259 

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter validation of object placement and size within the simulation volume, with clearer error reporting for invalid or unresolved bounds.
  * More robust handling of unresolved placement parameters when constraint resolution does not converge.

* **Tests**
  * Updated integration tests to construct objects explicitly (including colors and placement order).
  * Removed one redundant placement test.

* **Chores**
  * Eliminated obsolete internal test helpers.


ymahlau:
Hi @renaissancenerd, it is definitely good to remove the inner classes. Could you please also fix the failing test cases now?

ymahlau:
@renaissancenerd is this ready to merge? Please give me a short ping if yes

renaissancenerd:
> @renaissancenerd is this ready to merge? Please give me a short ping if yes

Yes, it is ready for merge

PR #305, opened 2026-04-27
feat: add CustomTimeSignalProfile and source signal visualization

JoGed:
Adds visualization of source signals in both the time and frequency domains, as well as functionality to define custom time-domain source profiles.

This serves to:

1. Verify the actual injected signal and its spectral content before running a simulation, e.g. to check that the source covers the frequency range used for response analysis.
2. Define custom time-domain source signals for direct source control, e.g. for adaptive spectral weighting in time-domain optimization workflows; see [Gedeon et al., IEEE TAP 2025](https://doi.org/10.1109/TAP.2025.3590211).

Plot example comparing a Gaussian pulse and a custom Hamming-windowed sinc signal:

source_comparison

## Summary by CodeRabbit

* **New Features**
  * Inject precomputed time-waveforms via a new custom temporal profile; enhanced temporal profiles can sample, compute one-sided spectra (optional normalization), and plot time+frequency with auto-scaled axes and file export.
  * Source objects expose convenient sampling, spectrum, and plotting helpers.

* **Documentation**
  * API docs updated to list the new temporal profile types and analysis/plotting methods.

* **Tests**
  * Extensive unit/integration tests for sampling, interpolation, spectra, plotting, and end-to-end physics comparisons.

### DIFF
--- a/src/fdtdx/objects/sources/profile.py
+++ b/src/fdtdx/objects/sources/profile.py
@@ -342,10 +342,17 @@ class CustomTimeSignalProfile(TemporalProfile):
     JIT before constructing this object.
 
     The sampled ``signal`` fully defines the injected time waveform, so the FDTD
-    loop does not apply an additional source-level ``phase_shift``. Optional
-    ``center_wave`` and ``fwidth`` values only describe the signal for reference
-    frequency and spectrum-plotting logic.
+    loop does not apply an additional source-level ``phase_shift``.
 
+    Unlike :class:`GaussianPulseProfile`, this profile does **not** accept a
+    ``center_wave`` or spectral-width parameter.  For an arbitrary waveform the
+    spectral centre is not a free parameter — it is an emergent property of the
+    signal.  Letting the user specify it separately would create a silent
+    inconsistency risk (e.g. labelling a broadband sinc pulse as if it were a
+    narrowband Gaussian).  Instead, :meth:`get_reference_frequency` computes the
+    power-weighted spectral centroid directly from ``signal``, and the frequency
+    axis of any plot is determined automatically from the actual spectrum content
+    via :func:`_auto_range`.
     """
 
     #: Pre-sampled waveform, shape ``(N,)``.  Lives in the pytree so JAX can
@@ -358,12 +365,6 @@ class CustomTimeSignalProfile(TemporalProfile):
     #: Simulation time at which ``signal[0]`` was sampled (seconds).
     start_time: float = frozen_field(default=0.0)
 
-    #: Optional center wavelength / frequency metadata for spectrum plots.
-    center_wave: WaveCharacter | None = frozen_field(default=None)
-
-    #: Optional spectral width metadata for spectrum plots.
-    fwidth: WaveCharacter | None = frozen_field(default=None)
-
     #: Interpolation mode: ``"linear"`` (default) or ``"nearest"``.
     interpolation: Literal["linear", "nearest"] = frozen_field(default="linear")
 
@@ -371,26 +372,20 @@ class CustomTimeSignalProfile(TemporalProfile):
     outside_value: float = frozen_field(default=0.0)
 
     def get_reference_frequency(self, period: float) -> float:
-        if self.center_wave is not None:
-            return self.center_wave.get_frequency()
-        return 1.0 / period
+        """Return the power-weighted spectral centroid of the stored signal.
 
-    def get_frequency_plot_range(
-        self,
-        period: float,
-        frequencies: np.ndarray,
-        spectrum: np.ndarray,
-    ) -> tuple[float, float] | None:
-        if self.center_wave is None or self.fwidth is None:
-            return None
-        del period, spectrum
-        f0 = self.center_wave.get_frequency()
-        df = self.fwidth.get_frequency()
-        lower = max(float(frequencies[0]), f0 - 4 * df)
-        upper = min(float(frequencies[-1]), f0 + 4 * df)
-        if upper <= lower:
-            return None
-        return lower, upper
+        The centroid is well-defined for any spectrum — it equals the peak
+        frequency for unimodal signals and the power-weighted average for
+        multi-band ones.  It is used by :func:`_auto_range` to anchor the
+        frequency axis of spectrum plots.
+        """
+        del period
+        frequencies = np.fft.rfftfreq(len(self.signal), d=self.time_step_duration)
+        spectrum = np.abs(np.fft.rfft(np.asarray(self.signal)))
+        total = float(np.sum(spectrum))
+        if total == 0.0:
+            return 0.0
+        return float(np.sum(frequencies * spectrum) / total)
 
     def get_amplitude(
         self,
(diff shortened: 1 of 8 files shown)


src/fdtdx/objects/sources/profile.py
```diff
@@ -1,13 +1,62 @@
...
+
+
+def _auto_range(
+    x: np.ndarray,
+    y: np.ndarray,
+    *,
+    relative_threshold: float,
+    pad_fraction: float,
+    center: float | None = None,
+) -> tuple[float, float]:
```
ymahlau, line 28: Could you please add a short explanation what these functions are supposed to achieve? this does not need to be a full docstring, but some explanation is necessary here. Same goes for function above.

src/fdtdx/objects/sources/profile.py
```diff
@@ -34,6 +83,160 @@ def get_amplitude(
         """
         raise NotImplementedError()
 
+    def get_reference_frequency(self, period: float) -> float:
+        """Return the frequency expected to dominate the plotted spectrum."""
+        return 1.0 / period
+
+    def get_time_plot_range(self, period: float, total_time: float) -> tuple[float, float] | None:
+        del period, total_time
+        return None
```
ymahlau, line 101: Are these functions supposed to be overriden by subclasses? If there is not functionality there then it would be best to raise a NotImplementedError here and add @abstracmethod decorator. Then you can also remove the None return type

src/fdtdx/objects/sources/profile.py
```diff
@@ -1,13 +1,62 @@
...
+    for label, scale in units:
+        if max_abs_value * scale >= 0.1:
+            return label, scale
+    return units[-1]
+
+
+def _auto_range(
+    x: np.ndarray,
+    y: np.ndarray,
+    *,
```
ymahlau: I generally dislike the use of star operator, please remove in the functions.

src/fdtdx/objects/sources/profile.py
```diff
@@ -34,6 +83,160 @@ def get_amplitude(
...
+        del period, total_time
+        return None
+
+    def get_frequency_plot_range(
+        self,
+        period: float,
+        frequencies: np.ndarray,
+        spectrum: np.ndarray,
+    ) -> tuple[float, float] | None:
+        del period, frequencies, spectrum
```
ymahlau, line 114: All of these functions are missing docstrings

src/fdtdx/objects/sources/profile.py
```diff
@@ -103,3 +331,97 @@ def get_amplitude(
...
+
+    Unlike :class:`GaussianPulseProfile`, this profile does **not** accept a
+    ``center_wave`` or spectral-width parameter.  For an arbitrary waveform the
+    spectral centre is not a free parameter — it is an emergent property of the
+    signal.  Letting the user specify it separately would create a silent
+    inconsistency risk (e.g. labelling a broadband sinc pulse as if it were a
+    narrowband Gaussian).  Instead, :meth:`get_reference_frequency` computes the
+    power-weighted spectral centroid directly from ``signal``, and the frequency
+    axis of any plot is determined automatically from the actual spectrum content
+    via :func:`_auto_range`.
```
ymahlau: Do you want to add a reference to your paper here?

src/fdtdx/objects/sources/profile.py
```diff
@@ -103,3 +331,97 @@ def get_amplitude(
...
+    """
+
+    #: Pre-sampled waveform, shape ``(N,)``.  Lives in the pytree so JAX can
+    #: differentiate through the interpolation if needed.
+    signal: jax.Array = field(on_setattr=[jnp.asarray])
```
ymahlau: the jnp.asarray as on setattr is not good, please remove

src/fdtdx/objects/sources/source.py
```diff
@@ -90,6 +91,55 @@ def place_on_grid(
         self = self._update_on_arrays()
         return self
 
+    def sample_time_signal(
+        self,
+        config: SimulationConfig | None = None,
+    ):
+        """Sample the source temporal profile with this source's phase and simulation cadence."""
```
ymahlau: Here are also better docstrings needed

src/fdtdx/objects/sources/source.py
```diff
@@ -90,6 +91,55 @@ def place_on_grid(
...
+        if config is None:
+            config = self._config
```
ymahlau: you should add a check here. What happens if self._config is also not set yet? Then the user should see an error that place_objects needs to be called before plotting or a config needs to be supplied.

tests/integration/fdtd/test_custom_time_signal_profile.py
```diff
@@ -0,0 +1,81 @@
+"""Integration tests for custom temporal profiles on placed sources."""
+
+from pathlib import Path
+
+import jax
+import jax.numpy as jnp
+import matplotlib
+
+matplotlib.use("Agg")
```
ymahlau: This line can be problematic because it also changes the behavior of other plotting tests that follow afterwards. That is a huge issue because the order of execution of tests should not change results

tests/unit/objects/sources/test_profile.py
```diff
@@ -1,9 +1,18 @@
...
 
+import math
+
+import jax
 import jax.numpy as jnp
+import matplotlib
+import numpy as np
+import pytest
+
+matplotlib.use("Agg")
```
ymahlau: same here.

tests/unit/objects/sources/test_profile.py
```diff
@@ -209,6 +218,251 @@ def test_get_amplitude_spectral_width_effect(self):
...
+        assert profile.start_time == 0.0
+        assert profile.interpolation == "linear"
+        assert profile.outside_value == 0.0
+
+    def test_accepts_array_like_signal(self):
+        """Test that Python array-likes are converted to JAX arrays."""
+        profile = CustomTimeSignalProfile(signal=[0.0, 1.0, 0.0], time_step_duration=0.25)
+
+        assert isinstance(profile.signal, jax.Array)
+        assert profile.signal.dtype == jnp.float32
```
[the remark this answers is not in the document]
ymahlau: I agree with ai review here

ymahlau:
Hi @JoGed thanks for your contribution! I left a number of comments above on what needs to be changed. In addition to these comments, your code coverage is only 72%. Please add more test cases to cover the lines of code you missed before (you can click on the report above to see which lines are covered or missed). 

As an optional suggestion, could you also add a small simulation test? For example a plane source with the custom source profile propagating through a layer of materials?

src/fdtdx/objects/sources/profile.py
JoGed: I added some docstrings here for clarification. I intentionally kept the methods as optional hooks returning `None`, as this triggers automatic plot-range detection for profiles without profile-specific time/frequency ranges (unlike e.g. the built-in Gaussian, where the center frequency and spectral width can define such ranges).

But I guess this is a design choice? If you prefer `@abstractmethod`s here, I will change it.

ymahlau: I think the design is good as is, thanks!

ymahlau:
Looks good to me now! Thanks for your first contribution!

PR #307, opened 2026-04-29
Propagate simulation dtype through sources and utility functions

teevee112:
Addresses https://github.com/ymahlau/fdtdx/issues/262

Replace hardcoded float32/complex64 in core utilities and source objects with dtype-aware parameters derived from SimulationConfig.dtype, so that float64 simulations run without JAX implicit-promotion warnings.

New dtype-propagation tests were evaluated but not included: JAX's `jax_enable_x64` is a process-global one-way flag that cannot be unset or scoped per-test, making float64 dtype assertions unreliable inside a shared pytest process.

Float64 correctness can be validated by running a simulation with JAX_ENABLE_X64=1 and SimulationConfig(dtype=jnp.float64) and confirming no JAX implicit-promotion warnings appear. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed numeric precision handling to ensure consistency across all simulation operations
  * Simulation configuration's precision setting now properly propagates through sources, detectors, mode calculations, and field computations
  * Eliminated implicit precision conversions during complex field and wave vector calculations

### DIFF
--- a/src/fdtdx/config.py
+++ b/src/fdtdx/config.py
@@ -105,6 +105,15 @@ class SimulationConfig(TreeClass):
             jax.config.update("jax_platform_name", "cpu")
 
     @property
+    def complex_dtype(self) -> jnp.dtype:
+        """Complex dtype matching the simulation's float dtype.
+
+        Returns:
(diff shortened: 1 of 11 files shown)


ymahlau:
Awesome @teevee112, many thanks for your first contribution! All changes look good to me and ready to merge!

PR #308, opened 2026-04-29
refactor: split FieldState from ArrayContainer and add reset() method

teevee112:
reset_array_container was a standalone function that manually zeroed E and H but silently omitted the PML auxiliary fields psi_E and psi_H, which were added later without updating the reset logic. This was the root cause of issue https://github.com/ymahlau/fdtdx/issues/260.

The fix introduces FieldState, a dedicated struct holding all dynamic fields that evolve each time step: E, H, psi_E, and psi_H. ArrayContainer now holds a single fields: FieldState instead of four flat fields, and a reset() method on ArrayContainer zeros the entire struct via jax.tree.map(jnp.zeros_like, self.fields). Adding a new dynamic field in the future — such as dispersion auxiliary fields — only requires adding it to FieldState.
The reset is then automatically correct with no code changes needed elsewhere, making this class of bug impossible to reintroduce.

The standalone reset_array_container function is removed; its three call sites in fdtd.py are replaced with arrays.reset().

## Summary by CodeRabbit

* **New Features**
  * `FieldState` is now publicly exported.
  * `ArrayContainer.reset()` added to reset field and detector/recording state.

* **Refactor**
  * Electromagnetic field data are now stored in a nested field container, with calling patterns and tests updated accordingly for consistency.

### DIFF
--- a/src/fdtdx/fdtd/container.py
+++ b/src/fdtdx/fdtd/container.py
@@ -316,7 +316,22 @@ class ArrayContainer(TreeClass):
         reset_detector_states: bool = True,
         reset_recording_state: bool = False,
     ) -> "ArrayContainer":
-        """Return a copy with all dynamic fields zeroed and material properties preserved."""
+        """Return a reset copy of this array container.
+
+        Dynamic field arrays are zeroed while material arrays and conductivity
+        arrays are preserved. Detector states are reset by default because they
+        accumulate time-dependent measurements. Recording state is preserved by
+        default so partial simulations can continue writing to the same buffers.
+
+        Args:
+            reset_detector_states: Whether to zero all detector state arrays.
+                Defaults to True.
+            reset_recording_state: Whether to zero recording data and state
+                arrays when a recording state is present. Defaults to False.
+
+        Returns:
+            A new ArrayContainer with reset dynamic state.
+        """
         arrays = self.aset("fields", jax.tree.map(jnp.zeros_like, self.fields))
 
         detector_states = self.detector_states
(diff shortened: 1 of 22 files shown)


ymahlau:
@coderabbitai resume

src/fdtdx/fdtd/container.py
```diff
@@ -298,47 +311,43 @@ class ArrayContainer(TreeClass):
...
+
+        return arrays
+
 
 # time step and arrays
 SimulationState = tuple[jax.Array, ArrayContainer]
 
 
+# Keep as a module-level alias so existing imports don't break during migration.
+# Will be removed once all call sites are updated.
```
ymahlau: please remove the old reset_array_container immediately, otherwise it will be forgotten about :)

src/fdtdx/fdtd/container.py
```diff
@@ -298,47 +311,43 @@ class ArrayContainer(TreeClass):
     #: field for magnetic conductivity terms. Defaults to None.
     magnetic_conductivity: jax.Array | None = None
 
+    def reset(
+        self,
+        reset_detector_states: bool = True,
+        reset_recording_state: bool = False,
+    ) -> "ArrayContainer":
+        """Return a copy with all dynamic fields zeroed and material properties preserved."""
```
ymahlau: please add a proper docstring here, similar to the old reset function

ymahlau:
relates to #294

ymahlau:
Thanks @teevee112, this is a good refactor! I left some minor comments above and some of the tests are still failing because of the new syntax. Could you resolve these last issues please? Afterwards this should be ready to merge

ymahlau (approved):

ymahlau:
Looks good to me! Many thanks @teevee112

PR #310, opened 2026-04-30
Provide progress callback for simulations

teevee112:
This PR adds a progress callback hook to FDTD execution so callers can observe simulation progress without depending directly on the built-in terminal progress bar.

This is useful for integrations where tqdm output is not the right interface, for example:

- notebooks or GUI applications that need to update their own progress display
- web/API jobs that need to report progress to a frontend or job queue
- tests or scripts that want deterministic progress events without parsing terminal output
- downstream libraries that wrap run_fdtd and need progress visibility without taking over stdout/stderr

The callback keeps progress reporting decoupled from presentation. The existing progress bar can remain the default user-facing behavior, while advanced callers can plug in their own reporting mechanism.

The default behaviour remains the same.

## Summary by CodeRabbit

* **New Features**
  * Added optional progress callback parameter to simulations. Users can supply a custom callback function to receive real-time progress updates (step index and total steps) during simulation runs.
  * Independent control over progress bar visibility, enabling flexible progress reporting without requiring UI-based indicators.

### DIFF
--- a/src/fdtdx/core/progress.py
+++ b/src/fdtdx/core/progress.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import math
+from collections.abc import Callable
 
 import jax
 import jax.experimental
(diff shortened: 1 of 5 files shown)


ymahlau:
@teevee112 thanks for this contribution also! I would like to test this PR in a bit more detail, which I will only get to do after my vacation, which is starting today. I will give a proper review sometime mid next week

ymahlau:
@coderabbitai review

ymahlau:
After some tests, PR looks good to me!

Issue #71, opened 2025-06-04
Dose fdtdx support non-uniform grids?


694856790:
I suggest that the data format should follow  non-uniform grids, even if it is now  uniform grids. Uniform grids are a special case of non-uniform grids, and if the data structure is not well done at first, it will become more difficult later. Obviously MEEP is in this dilemma.

ymahlau:
To answer the question in the title, currently fdtdx does not support non-uniform grids. I think in the near to mid future, I will also not have time to implement this feature. 

What exactly do you mean by data format? 

It is only possible to implement partial non-uniform grids for FDTD. Specifically, the grid still needs to have a grid structure made from cuboid grid cells, but the size of the cuboids can vary throughout the simulation volume. For truly non-uniform meshes of arbitrary structures, the FDTD update equations don't work anymore. In that case you will need to perform FEM-Analysis. If you are interested in this, [jax-fem](https://github.com/deepmodeling/jax-fem/tree/main) might be a good starting point.

To implement non-uniform grids, one would need to specify three 1d-arrays specifying how one array index step corresponds to metrical units in the direction of this axis. Then the update equations would need to be adjusted and also some physics computations like modes, phasors need to be adjusted.

Additionally, special care needs to be taken to adjust the object placement API. We should keep the easy placement of objects and also implement an intuitive way specify the grid sizes of the non-uniform grid. One way to do this might be to specify a uniform-resolution for every object individually and then automatically calculate the non-uniform grid for the whole simulation volume.

694856790:
> 回答标题中的问题，目前 fdtdx 不支持非均匀网格。我觉得在近期或中期，我也没有时间实现这个功能。
> 
> 数据格式到底是什么意思？
> 
> FDTD 只能实现部分非均匀网格。具体来说，网格仍然需要由长方体网格单元构成的网格结构，但长方体的尺寸可以在整个模拟体积内变化。对于任意结构的真正非均匀网格，FDTD 更新方程不再适用。在这种情况下，您需要进行 FEM 分析。如果您对此感兴趣，[jax-fem](https://github.com/deepmodeling/jax-fem/tree/main)或许是一个不错的起点。
> 
> 要实现非均匀网格，需要指定三个一维数组，分别指定每个数组索引步长对应该轴方向上的度量单位。然后，需要调整更新方程，并调整一些物理计算，例如模态、相量等。
> 
[...]

PR #312, opened 2026-05-04
Add rectilinear grid support with one solver-facing grid path

teevee112:
Addresses https://github.com/ymahlau/fdtdx/issues/71

## Summary

This PR introduces rectilinear grid support and removes the scalar `SimulationConfig.resolution` path. The main design choice is that **uniform grids are treated as a special case of rectilinear grids**, not as a separate solver mode.

The solver-facing representation is now `RectilinearGrid`, which stores explicit physical edge coordinates:

- `x_edges`
- `y_edges`
- `z_edges`

Uniform grids enter through `UniformGrid(spacing=...)`, which is a lightweight grid policy. During placement, it resolves to a concrete `RectilinearGrid`. After that point, simulation code consumes one realized grid representation.

## Philosophy

The goal is to avoid having two competing discretization models:

- one scalar-resolution path for uniform grids
- one metric-aware path for non-uniform grids

That split would make future work fragile because new code could accidentally keep using scalar spacing assumptions. Instead, this PR makes the grid itself the source of truth for physical metrics.

Uniform grids still stay efficient and simple, but they now follow the same API and data model as non-uniform grids. This should make future grid work easier to reason about and easier to test.

### One realized solver grid

`RectilinearGrid` is the concrete grid used by solver internals. It owns:

- edge coordinates
- cell widths
- cell centers
- physical extents
- face areas
- cell volumes
- CFL time-step calculation
- coordinate-to-index snapping helpers

This keeps placement, detectors, sources, PML, exports, and update equations from each inventing their own grid math.

### Separate grid policy from realized grid

`UniformGrid(spacing=...)` represents user intent before the final simulation shape is known.

It resolves to:

```python
RectilinearGrid.uniform(shape=..., spacing=...)
```
during setup.

This leaves room for future policy objects such as AutoGrid, QuasiUniformGrid, refinement overrides, or wavelength-aware meshing, while keeping the solver backend unchanged.

### No backwards-compatible scalar resolution path

This PR intentionally removes SimulationConfig.resolution.

That is a breaking API change, but it avoids maintaining two sources of truth. The replacement is explicit:
```
SimulationConfig(
    time=...,
    grid=UniformGrid(spacing=50e-9),
)
```
or:
```
SimulationConfig(
    time=...,
    grid=RectilinearGrid.custom(
        x_edges=...,
        y_edges=...,
        z_edges=...,
    ),
)
```

### Physical reductions by default

Reduced detector outputs now use physical weights:

- energy reductions use cell volumes
- Poynting flux reductions use face areas
- mode overlap uses detector-plane face areas

The old raw-sum behavior was grid-dependent and becomes misleading on non-uniform grids. A refined region should not change total energy or flux just because it contains more cells.

## What Changed

- Added UniformGrid as the uniform-grid policy
- Made RectilinearGrid the concrete solver-facing grid model
- Removed SimulationConfig.resolution
- Updated placement to resolve grid policy before objects are placed
- Updated real-coordinate and real-shape placement to use grid edge coordinates
- Added local metric handling for Yee curls and CFL calculation
- Updated field interpolation for stretched cells
- Updated PML grading and Bloch phase calculations to use physical distances
- Updated sources and mode solver wrappers to pass/use physical coordinates where available
- Updated detector reductions to use area/volume weights
- Updated geometry masks for spheres, cylinders, and polygons to sample physical cell centers
- Added VTR export for rectilinear grids and kept VTI limited to uniform image-data grids
- Updated examples, checks, tests, and docs-facing API usage to the new grid model

### Breaking API change

Removing SimulationConfig.resolution means existing scripts must change to grid=UniformGrid(spacing=...).

This is intentional. Keeping resolution as a compatibility alias would preserve an attractive nuisance: new code could continue depending on scalar spacing and silently break non-uniform behavior.

### More metric plumbing

Some code now asks the grid for physical areas, volumes, centers, and extents. This is more explicit than scalar arithmetic, but it makes assumptions visible and testable.

### Performance considerations

Metric-aware code can allocate more intermediate arrays, especially for broadcasted curl scales, area weights, volume weights, and source coordinate grids.

This PR keeps the implementation correct and modular first. Follow-up optimization opportunities include:

- caching common solver metric arrays
- using separable 1D weights where possible instead of full 3D arrays
- precomputing source/detector metric coordinates at placement time
- profiling JIT-compiled update loops for metric broadcast overhead

## Validation

Pre-commit passes:

uv run pre-commit run --all-files

The branch also includes focused tests for:

- uniform and non-uniform grid construction
- coordinate snapping and placement behavior
- CFL calculation on stretched grids
- local curl metric accuracy
- weighted energy / flux / mode-overlap reductions
- PML physical-depth profiles
- Bloch phase physical lengths
- geometry masks on rectilinear grids
- source profile behavior on stretched grids
- VTR/VTI export behavior
- full FDTD run on a mildly stretched grid matches uniform-grid observables

## Future Work

This PR does not implement automatic meshing.

The intended future direction is to add grid policy objects that resolve to RectilinearGrid, for example:

- AutoGrid
- QuasiUniformGrid
- wavelength-aware refinement
- material-interface refinement
- user mesh override regions

Those policies should stay outside the solver. The solver should continue consuming realized RectilinearGrid coordinates.

## Summary by CodeRabbit

* **New Features**
  * Added non-uniform (rectilinear) grid support for FDTD simulations, enabling stretched mesh geometries for improved computational efficiency.
  * Introduced grid-based simulation configuration API using `UniformGrid` and `RectilinearGrid` classes.
  * Added VTK RectilinearGrid export (`export_vtr`) for non-uniform grid data.
  * Enhanced detector plotting and mode solving for non-uniform grids.

* **API Changes**
  * `SimulationConfig` now uses `grid=UniformGrid(spacing=...)` instead of `resolution=...`.


ymahlau:
Hi @teevee112, awesome that you started tackling this! Without looking at the code, I agree with the design decisions, removing `resolution` is a good way to go. Will give a proper review later this day.

src/fdtdx/fdtd/initialization.py
```diff
@@ -742,6 +762,66 @@ def _center_to_bounds(real_pos: float, resolution: float, size: int) -> tuple[in
...
+    """Convert a physical length to a grid-cell count.
+
+    For uniform grids this intentionally uses the historical ``round(length/dx)``
+    rule.  For non-uniform grids it chooses enough cells from the lower domain
+    edge to cover the requested metric length.
+    """
+    if _has_nonuniform_grid(config):
+        assert config.grid is not None
+        return config.grid.length_to_cell_count(axis, length, snap="upper")
+    return round(length / config.require_uniform_grid())
```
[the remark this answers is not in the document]
teevee112: This is a known limitation of metric-size placement on non-uniform grids. The current behavior is intentional for this PR — proper position-aware sizing would require a larger redesign of the placement system and is out of scope here.

examples/width_sweep_analysis.py
```diff
@@ -200,7 +200,7 @@ def make_waveguide_simulation(wavelength, mesh, width, height, absorber_thicknes
 
     # Define simulation configuration
     config = fdtdx.SimulationConfig(
-        resolution=resolution,
+        grid=fdtdx.UniformGrid(spacing=resolution),
         time=time,
         courant_factor=courant_factor,
     )
```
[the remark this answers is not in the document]
teevee112: Fixed. Both object._config.resolution calls updated to object._config.require_uniform_grid()

src/fdtdx/fdtd/initialization.py
```diff
@@ -742,6 +762,66 @@ def _center_to_bounds(real_pos: float, resolution: float, size: int) -> tuple[in
...
+        return config.grid.coord_to_index(axis, coord, snap="nearest")
+    return round(coord / config.require_uniform_grid())
+
+
+def _center_to_bounds_for_grid(config: SimulationConfig, axis: int, real_pos: float, size: int) -> tuple[int, int]:
+    """Convert a physical center and resolved grid size to edge bounds."""
+    if _has_nonuniform_grid(config):
+        assert config.grid is not None
+        return config.grid.bounds_for_center(axis, real_pos, size)
+    return _center_to_bounds(real_pos, config.require_uniform_grid(), size)
```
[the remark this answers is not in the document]
teevee112: Valid in principle, but all current UniformGrid usage sets origin to (0, 0, 0) (the default), so there's no practical regression. Routing the uniform path through config.grid.coord_to_index() for correctness can be a follow-up.

src/fdtdx/objects/boundaries/perfectly_matched_layer.py
```diff
@@ -166,6 +182,39 @@ def _compute_pml_profile(
...
+        else:
+            interface = edges[0]
+            dE = edges[:-1] - interface
+            dH = centers - interface
+            dH = dH.at[0].set(0)
+
+        # If a boundary cell has zero width validation would already have failed
+        # in RectilinearGrid, but this keeps the normalization robust for diagnostics.
+        del widths
+        return dE, dH, norm
```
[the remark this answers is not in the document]
teevee112: The asymmetry between + and - is intentional — on a non-uniform grid, edge-based depths are the natural definition since there's no uniform cell size to center-offset by. A parity test against the uniform limit is a reasonable suggestion and can be added in a follow-up, but the current behavior is not a regression

src/fdtdx/objects/detectors/diffractive.py
```diff
@@ -100,6 +100,48 @@ def _shape_dtype_single_time_step(self) -> dict[str, jax.ShapeDtypeStruct]:
...
+
+        def interp_y(component):
+            return jax.vmap(lambda row: jnp.interp(target_y, y_centers, row))(component)
+
+        field = jax.vmap(interp_x)(field)
+        field = jax.vmap(interp_y)(field)
+
+        dx = float(target_x[1] - target_x[0]) if nx > 1 else 1.0
+        dy = float(target_y[1] - target_y[0]) if ny > 1 else 1.0
+        return field, dx, dy
```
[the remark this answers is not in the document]
teevee112: Fixed — removed the float() calls. target_x and target_y are built from jnp.linspace over concrete grid data, so today they aren't traced values and float() happens to work. But if this method is ever called in a context where grid centers become traced (e.g., a future differentiable grid), it would raise TracerArrayConversionError. Dropped the casts and updated the return type annotation to jax.Array | float accordingly.

src/fdtdx/objects/detectors/mode.py
```diff
@@ -87,6 +89,56 @@ def place_on_grid(
...
+        The propagation axis is the plane normal.  For legacy construction paths
+        without an explicit ``RectilinearGrid``, the scalar grid spacing supplies the
+        uniform face area.
+        """
+        grid = self._config.realized_grid
+        if grid is not None:
+            return grid.face_area(axis=self.propagation_axis, slice_tuple=self.grid_slice_tuple)
+
+        spacing = self._config.require_uniform_grid()
+        return jnp.ones(self.grid_shape, dtype=self.dtype) * spacing * spacing
```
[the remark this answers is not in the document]
teevee112: Fixed — changed to jnp.float32. Face areas are purely real geometric quantities; allocating them as complex64 (from self.dtype) was wasteful.

src/fdtdx/objects/device/device.py
```diff
@@ -59,9 +63,13 @@ def matrix_voxel_grid_shape(self) -> tuple[int, int, int]:
...
-            round(self.grid_shape[1] / self.single_voxel_grid_shape[1]),
-            round(self.grid_shape[2] / self.single_voxel_grid_shape[2]),
+            self._matrix_voxel_grid_shape_override
+            if self._matrix_voxel_grid_shape_override != INVALID_SHAPE_3D
+            else (
+                round(self.grid_shape[0] / self.single_voxel_grid_shape[0]),
+                round(self.grid_shape[1] / self.single_voxel_grid_shape[1]),
+                round(self.grid_shape[2] / self.single_voxel_grid_shape[2]),
+            )
         )
```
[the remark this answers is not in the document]
teevee112: Existing behavior predating this PR. There is already a downstream divisibility check at line 192 that catches actual misalignments and raises. Not addressing in this PR

src/fdtdx/objects/device/device.py
```diff
@@ -121,21 +161,35 @@ def place_on_grid(
             if partial_grid is not None:
                 voxel_grid_shape.append(partial_grid)
             elif partial_real is not None:
-                voxel_grid_shape.append(round(partial_real / config.resolution))
+                if uses_physical_design_grid:
+                    if partial_real <= 0:
+                        raise ValueError(f"Physical design voxel size must be positive for {axis=}.")
+                    physical_design_shape.append(float(partial_real))
+                    matrix_shape_override.append(max(1, math.ceil(self.real_shape[axis] / partial_real)))
+                    voxel_grid_shape.append(1)
```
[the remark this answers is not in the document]
teevee112: Existing behavior predating this PR. There is already a downstream divisibility check at line 192 that catches actual misalignments and raises. Not addressing in this PR

src/fdtdx/core/grid.py
```diff
@@ -4,6 +4,423 @@
...
+    physical cell spacing while the final simulation shape may still be unknown.
+    Object placement resolves this policy to a concrete ``RectilinearGrid`` once
+    the volume shape is known.
+
+    Keeping uniform spacing here avoids a second scalar discretization source on
+    ``SimulationConfig``.  Uniform grids and explicitly non-uniform grids both
+    enter the solver through the same realized ``RectilinearGrid`` structure.
+    """
+
+    spacing: float = frozen_field()
```
ymahlau, line 25: minor comment: please add docstrings to all user-facing attributes or at best to all of them. This also applies to the other new classes introduced in this PR

src/fdtdx/config.py
```diff
@@ -118,19 +124,67 @@ def courant_number(self) -> float:
         """
         return self.courant_factor / math.sqrt(3)
 
+    def require_grid(self, shape: tuple[int, int, int] | None = None) -> RectilinearGrid:
```
ymahlau: I dont like the require ... name here, maybe get_resolved_grid fits better?

ymahlau: or just rectilinear_grid?

src/fdtdx/config.py
```diff
@@ -118,19 +124,67 @@ def courant_number(self) -> float:
...
+            A concrete ``RectilinearGrid``.
+        """
+        if isinstance(self.grid, RectilinearGrid):
+            return self.grid
+        if shape is None:
+            raise ValueError("A grid shape is required to resolve UniformGrid.")
+        return self.grid.resolve(shape)
+
+    @property
+    def realized_grid(self) -> RectilinearGrid | None:
```
ymahlau: here the name is also a bit misleading as well as the docstring

src/fdtdx/config.py
```diff
@@ -118,19 +124,67 @@ def courant_number(self) -> float:
...
+        return None
+
+    @property
+    def has_nonuniform_grid(self) -> bool:
+        """Whether the realized solver grid is non-uniform."""
+        grid = self.realized_grid
+        return grid is not None and not grid.is_uniform
+
+    def require_uniform_grid(self) -> float:
+        """Return uniform spacing for paths that require a uniform grid.
```
ymahlau: see above

tests/integration/conversion/test_vti.py
```diff
@@ -47,5 +47,5 @@ def test_export_arrays_snapshot_to_vti(tmp_path):
     assert 'Name="H"' in content
 
     # Check spacing/resolution was passed correctly
-    res_str = f"{config.resolution} {config.resolution} {config.resolution}"
+    res_str = f"{config.require_uniform_grid()} {config.require_uniform_grid()} {config.require_uniform_grid()}"
```
ymahlau: It would be good to add test cases for the new vtr export function here

src/fdtdx/core/physics/modes.py
```diff
@@ -118,8 +119,8 @@ def compute_mode(
         inv_permittivities (jax.Array): 3D array of inverse relative permittivity values
         inv_permeabilities (jax.Array | float): 3D array of inverse relative permittivity values or single float for
             uniform permeability distribution.
-        resolution (float): resolution of the simulation grid in meter. For example a grid spacing of 10nm should be
-            given as 10e-9.
+        resolution (float): Uniform-grid spacing in metres. For example a grid spacing of 10nm should be given as
```
ymahlau: I think it would make sense to declare resolution as optional with None default value here, if coords are given it should not be needed

src/fdtdx/core/grid.py
```diff
@@ -4,6 +4,423 @@
...
+    widths, centers, face areas, and volumes are derived from these arrays.
+
+    Notes:
+        This class intentionally does not encode automatic mesh generation
+        policy.  Future policy objects such as ``AutoGrid`` or
+        ``QuasiUniformGrid`` should resolve to ``RectilinearGrid`` before the
+        solver runs.
+    """
+
+    x_edges: jax.Array = frozen_field()
```
ymahlau: Usually, jax arrays should not be declared as frozen-fields. If they are declared as frozen, then it will no longer be possible to use the data structure as non-static argument in a jitted function

src/fdtdx/core/grid.py
```diff
@@ -4,6 +4,423 @@
...
+            lower, upper = slice_tuple[transverse_axis]
+            widths.append(self.cell_widths(transverse_axis)[lower:upper])
+        area_2d = widths[0][:, None] * widths[1][None, :]
+        shape = [1, 1, 1]
+        shape[transverse_axes[0]] = area_2d.shape[0]
+        shape[transverse_axes[1]] = area_2d.shape[1]
+        return area_2d.reshape(shape)
+
+    def cell_volume(self, slice_tuple: tuple[tuple[int, int], tuple[int, int], tuple[int, int]]) -> jax.Array:
+        """Return per-cell volume weights broadcast to a 3D slice shape."""
```
ymahlau, line 426: I would strongly advise against ever materialising a 3d array of values for the grid. For multi-gpu this array can easily be larger than the vram of a single gpu and crash the simulation since it would need distributed initialization. Please remove this function and use the separated per-axis arrays instead.

src/fdtdx/core/grid.py
```diff
@@ -47,35 +466,71 @@ def calculate_time_offset_yee(
...
-    )
-    xyz_H = jnp.stack(
-        [
-            xyz + jnp.asarray([0, 0.5, 0.5])[None, None, None, :],
-            xyz + jnp.asarray([0.5, 0, 0.5])[None, None, None, :],
-            xyz + jnp.asarray([0.5, 0.5, 0])[None, None, None, :],
-        ]
-    )
+    if coordinate_edges is None:
+        # phase variation in grid-index units; convert to metres at the end
```
ymahlau: Is it necessary to have two different code paths here? I would suggest whenever it does not gain much performance increase to use a single code path with the generalized non-uniform grid to make the code less convoluted. Would also be okay to fix later if you dont want to address in this PR

src/fdtdx/core/grid.py
```diff
@@ -211,3 +666,25 @@ def polygon_to_mask(
...
 
     return mask
+
+
+def polygon_to_mask_at_points(
+    x_coords: np.ndarray,
+    y_coords: np.ndarray,
+    polygon_vertices: np.ndarray,
+) -> np.ndarray:
+    """Generate a 2D polygon mask at explicit sample coordinates.
```
ymahlau, line 645: Same here: can avoid duplicate code by re-using function above

src/fdtdx/fdtd/initialization.py
```diff
@@ -742,6 +762,66 @@ def _center_to_bounds(real_pos: float, resolution: float, size: int) -> tuple[in
     return lower, upper
 
 
+def _has_nonuniform_grid(config: SimulationConfig) -> bool:
+    """Return True when placement should use physical edge coordinates.
```
ymahlau: This is probably better as a getter on the config itself. It is also not very good to adjust the source code to fit the tests, might be better to adjust the tests instead

ymahlau: Same for all of the other helper functions below

ymahlau:
As a general comment: A lot of the code introduced in this PR has two code paths: one for uniform and one for non-uniform grids. I would like to see only one code path if it does not lead to large performance decreases, because that makes the code much easier to maintain. It is okay if you dont want to fix this now, but in that case please create a seprate issue to track it.

src/fdtdx/fdtd/update.py
```diff
@@ -103,8 +105,16 @@ def pad_fields_for_boundaries(
     boundaries = objects.boundary_objects
     if boundaries:
         volume_shape = objects.volume.grid_shape
+        grid = getattr(config, "grid", None)
+        if grid is not None and not grid.is_uniform:
+            # ``apply_pad_correction`` still has a legacy scalar-resolution
```
ymahlau: might make sense to remove then

src/fdtdx/objects/boundaries/perfectly_matched_layer.py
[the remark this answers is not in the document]
ymahlau: agree with coderabbit, remove dead code

src/fdtdx/config.py
```diff
@@ -51,8 +52,13 @@ class SimulationConfig(TreeClass):
...
 
-    #: Spatial resolution of the simulation grid in meters.
-    resolution: float = frozen_field()
+    #: Spatial grid configuration.
+    #:
+    #: ``UniformGrid`` is an unresolved policy used while the final volume shape
+    #: is still being inferred.  ``RectilinearGrid`` is the realized solver grid
+    #: with explicit physical edge coordinates.  Placement resolves policies to
+    #: ``RectilinearGrid`` so compiled FDTD code has exactly one metric source.
+    grid: UniformGrid | RectilinearGrid = frozen_field()
```
ymahlau: A cleaner implementation might be to have a private attribute _resolved_grid which is always a rectilinear grid computed in post-init. For all subsequent computations this grid should then be used.

ymahlau:
As another general comment: usually its best practice to keep jax.Array attributes non-frozen as well as any nested pytreeclass (for example a simulation config inside another class should not be frozen)

src/fdtdx/core/grid.py
```diff
@@ -4,6 +4,423 @@
...
+        return lower_coord + 0.5 * (position + 1.0) * (upper_coord - lower_coord)
+
+    def bounds_for_anchor(self, axis: int, size: int, anchor: float, position: float) -> tuple[int, int]:
+        """Choose a uniform-grid interval from an object anchor."""
+        anchor_cell = self.coord_to_index(axis, anchor, snap="nearest")
+        offset = round(0.5 * (position + 1.0) * size)
+        lower = anchor_cell - offset
+        return lower, lower + size
+
+    def cell_volume(self, slice_tuple: tuple[tuple[int, int], tuple[int, int], tuple[int, int]]) -> jax.Array:
```
ymahlau, line 104: I believe that most of these helper functions are not really necessary and just duplicates of the rectilinear grid below. If possible remove and just always use the resolved grid to make the code easier to maintain.

src/fdtdx/objects/detectors/detector.py
```diff
@@ -80,6 +81,94 @@ def num_time_steps_recorded(self) -> int:
...
+        assuming equal-volume voxels.  A scalar-resolution fallback preserves
+        compatibility for tests and older construction paths without an explicit
+        ``RectilinearGrid``.
+        """
+        grid = self._config.realized_grid
+        if grid is not None:
+            return grid.cell_volume(self.grid_slice_tuple)
+
+        spacing = self._config.require_uniform_grid()
+        return jnp.ones(self.grid_shape, dtype=self.dtype) * spacing * spacing * spacing
```
ymahlau: minor comment: maybe just use spacing ** 3 for a bit cleaner code :D

ymahlau:
As another general comment, it is generally not a good idea to use @cached_property when working with jax. It is better to use a private attribute which is initialized in post-init. 

I am working in the drinx repo on a more convenient way to declare cached values, but will take some time.

src/fdtdx/objects/detectors/diffractive.py
```diff
@@ -100,6 +100,48 @@ def _shape_dtype_single_time_step(self) -> dict[str, jax.ShapeDtypeStruct]:
...
 
+    def _transverse_centers(self, plane_dims: list[int]) -> tuple[jax.Array, jax.Array] | None:
+        """Return physical transverse cell centers for the detector plane.
+
+        FFT order decomposition requires equally spaced samples.  On rectilinear
+        non-uniform grids, fields are first resampled onto a uniform grid spanning
+        the same detector-center extent.  Uniform grids keep the legacy direct
+        FFT path.
+        """
+        if not self._config.has_nonuniform_grid:
```
ymahlau, line 112: not sure, but i think this is duplicate code

src/fdtdx/objects/detectors/diffractive.py
```diff
@@ -100,6 +100,48 @@ def _shape_dtype_single_time_step(self) -> dict[str, jax.ShapeDtypeStruct]:
...
+            centers.append(grid.centers(axis)[lower:upper])
+        return tuple(centers)  # type: ignore[return-value]
+
+    @staticmethod
+    def _uniform_resample_2d(
+        field: jax.Array,
+        x_centers: jax.Array,
+        y_centers: jax.Array,
+    ) -> tuple[jax.Array, jax.Array | float, jax.Array | float]:
+        """Resample ``(component, x, y)`` fields onto uniform transverse centers."""
```
ymahlau: This is probably also a function that can be refactored into src/core

src/fdtdx/objects/device/device.py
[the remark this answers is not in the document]
ymahlau: yes, please open a github issue

src/fdtdx/core/grid.py
teevee112: Done — added `#:` docstrings to all user-facing attributes on both `UniformGrid` and `RectilinearGrid`.

src/fdtdx/config.py
teevee112: Renamed to `resolve_grid` (method) and `resolved_grid` (property). Used verb/past-participle to distinguish the callable from the property. Updated all call sites.

src/fdtdx/config.py
teevee112: Renamed `realized_grid` → `resolved_grid` and updated the docstring to clarify it returns `None` when the grid is still an unresolved `UniformGrid`.

src/fdtdx/config.py
teevee112: Renamed `require_uniform_grid` → `uniform_spacing` — the new name describes what it returns rather than what it requires.

tests/integration/conversion/test_vti.py
teevee112: Added `test_export_vtr_uniform_grid` and `test_export_vtr_nonuniform_grid` in the same file.

src/fdtdx/core/physics/modes.py
teevee112: Fixed — `resolution` is now `float | None = None`. A `ValueError` is raised when `transverse_coords is None` and `resolution is None`.

src/fdtdx/core/grid.py
teevee112: Fixed — `x_edges`, `y_edges`, `z_edges` changed from `frozen_field()` to `field()`. Also changed `SimulationConfig.grid` for the same reason.

src/fdtdx/core/grid.py
teevee112: The implementation uses JAX broadcasting: `dx[x0:x1, None, None] * dy[None, y0:y1, None] * dz[None, None, z0:z1]`. No full-grid array is pre-allocated or stored — the result is a local detector-slice-sized array materialised on demand. Happy to change the interface to return per-axis 1D arrays if you'd prefer to be explicit about this.

src/fdtdx/core/grid.py
teevee112: Agreed — will address in a follow-up PR. The uniform/non-uniform split in the Bloch phase calculation still exists; leaving it here to avoid further scope creep.

src/fdtdx/core/grid.py
teevee112: Fixed — `polygon_to_mask` now delegates to `polygon_to_mask_at_points` instead of duplicating the meshgrid logic.

src/fdtdx/fdtd/initialization.py
teevee112: Fixed — `_has_nonuniform_grid` removed, replaced with a `has_nonuniform_grid` property on `SimulationConfig` itself.

teevee112: Fixed — all placement helper functions now call `config.grid.method()` directly. The only remaining helper is `_real_length_to_grid_size`, which centralises the `snap` mode selection (upper for non-uniform, nearest for uniform).

src/fdtdx/fdtd/update.py
teevee112: Fixed — replaced `getattr(config, 'grid', None)` with `config.has_nonuniform_grid`.

src/fdtdx/objects/boundaries/perfectly_matched_layer.py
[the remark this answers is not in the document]
teevee112: Fixed — removed the leftover `widths = edges[1:] - edges[:-1]` and `del widths` lines.

src/fdtdx/objects/detectors/detector.py
teevee112: Fixed — now uses `spacing**3`.

src/fdtdx/config.py
teevee112: The challenge with an eager `_resolved_grid` in `__post_init__` is that `UniformGrid` can't resolve until the simulation volume shape is known, which only becomes available during `place_objects`. The current `resolved_grid` property returns `None` until then and callers fall back to `uniform_spacing()`. Open to revisiting if you have a design that doesn't require shape at init time.

src/fdtdx/core/grid.py
teevee112: The `UniformGrid` methods are needed during pre-placement (before shape is known), where `RectilinearGrid` isn't available yet. They implement the same interface with simpler scalar math. If `UniformGrid` is always resolved eagerly at some earlier point, these could be removed — ties into the `_resolved_grid` design discussion above.

src/fdtdx/objects/detectors/diffractive.py
teevee112: The two arms handle genuinely different cases: uniform grids go through a direct FFT path while non-uniform grids must first resample onto a uniform grid. They share the final FFT call but the setup is different enough that merging them would obscure intent. Open to suggestions.

src/fdtdx/objects/detectors/diffractive.py
teevee112: Agreed, could live in `core/physics`. Leaving for a follow-up to keep this PR focused.

src/fdtdx/objects/device/device.py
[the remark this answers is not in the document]
teevee112: The voxel snap-to-grid behavior on non-uniform grids is a known limitation. Noted as a follow-up; not opening a separate issue to avoid scope creep here.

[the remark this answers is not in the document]
teevee112: raised issue here: https://github.com/ymahlau/fdtdx/issues/322

teevee112:
TODOs in a follow up PR:

Bloch phase unification                                                                                                                                                                          
  Where: src/fdtdx/core/grid.py, calculate_bloch_phase, the if coordinate_edges is None: branch
  Unify the two code paths here — the uniform path (grid-index space) is a special case of the non-uniform path and can be removed once RectilinearGrid is always available at this point.    
   
  ---                                                                                                                                                                                                 
Device physical voxels                                                 
  Where: src/fdtdx/objects/device/device.py, matrix_voxel_grid_shape property, lines 69–72                                                                                                          
  Add proper support for physical-unit design voxels on non-uniform grids. Currently round(grid_shape / single_voxel_grid_shape) assumes uniform cell sizes. Needs a resampling layer or 
  snapping to RectilinearGrid cell boundaries.                                                                                                                                                        
                                                                                                                                                                                                      
  ---                                                                                                                                                                                                 
_uniform_resample_2d refactor                                                                                                                                                                    
  Where: src/fdtdx/objects/detectors/diffractive.py, _uniform_resample_2d static method
  Move this into src/fdtdx/core/physics — it's a general utility not specific to this detector.

src/fdtdx/objects/detectors/diffractive.py
ymahlau: I see, I was looking at so much code during review that I just thought I had seen this before somewhere :) In that case just leave it as is, this genuinely needs two different code paths

ymahlau:
@teevee112 It is a bit difficult for me to keep an overview here of what you changed, what you still intend to change and what you want to refer into a different PR. Could you give me a brief overview and tag me as soon as you think the PR might be ready?

src/fdtdx/objects/detectors/diffractive.py
teevee112: implemented

src/fdtdx/core/grid.py
teevee112: fixed

teevee112:
All review comments addressed. Summary of changes made in response to feedback:

Renamed require_uniform_grid → rectilinear_grid and realized_grid → resolved_grid; updated all call sites and docstrings
Added docstrings to all user-facing attributes on new classes in grid.py and config.py
Added VTI export test cases for non-uniform grids in tests/integration/conversion/test_vti.py
Fixed VTI export: origin now uses actual grid coordinates instead of hardcoded "0 0 0"
Fixed frozen fields: JAX array attributes and nested pytrees now use field() instead of frozen_field()
Replaced @cached_property with private attributes initialized in post_init
Fixed stale comment in test_stop_conditions.py (resolution= → grid=UniformGrid(...))
Fixed PoyntingFluxDetector crash when propagation axis is ambiguous
Unified Bloch phase code paths: single physical-space path in calculate_time_offset_yee
Moved resample_to_uniform_2d from DiffractiveDetector into src/fdtdx/core/physics/metrics.py as a general utility
Physical-unit design voxels (partial_voxel_real_shape) on non-uniform grids now raise NotImplementedError with an actionable message; deferred to a follow-up PR

Deferred task:

https://github.com/ymahlau/fdtdx/issues/322

ymahlau:
Again, thanks for the implementation!

Issue #329, opened 2026-05-11
apply_params currently not working with mode source / detector


ymahlau:
Currently, the apply_params method fails, because of tracer leakage (see image attached).

<img width="2272" height="715" alt="Image" src="https://github.com/user-attachments/assets/56c15378-3ece-4077-902f-f7163f3c0ec3" />

ymahlau:
@teevee112 this was introduced in #312, could you please take a look and fix ASAP? If you dont have time please let me know, because we cannot leave this bug in the main branch like this, because it basically removes the capability to use any mode detector or source

ymahlau:
We should add an integration test for this specifically (apply params / mode solver called under jit)

PR #331, opened 2026-05-12
prevent TracerArrayConversionError in mode solver under jax.jit

teevee112:
Addresses  https://github.com/ymahlau/fdtdx/issues/329

- All _local_edge_coordinates() implementations now return np.ndarray via edges_np() instead of jax.Array via edges().
- The uniform-grid path in calculate_time_offset_yee switches from jnp.arange to np.arange, so both branches of resolved_edges are the same type. This removes the union from coordinate_edges and
component_positions.
- Cell centers derived from numpy edges are explicitly wrapped in jnp.asarray() before being passed to _linear_interpolate_rectilinear_2d, which indexes them with JAX-traced integers from
jnp.searchsorted.
- compute_mode's transverse_coords is narrowed from Sequence[ArrayLike] to Sequence[np.ndarray], so ty check statically rejects passing a jax.Array there.

Why

Grid edge coordinates are static metadata — they never change during an optimization run and are never differentiated through. Storing them as jax.Array pytree leaves was incidental, not intentional.
The consistent rule is now: edge coordinates flow as numpy up to the point where they enter a JAX computation, at which point they are explicitly converted.

  Test plan

  - TestModePlaneSourceJitSafety::test_local_edge_coordinates_returns_numpy_with_rectilinear_grid — asserts _local_edge_coordinates() returns np.ndarray on a rectilinear config
  - TestModePlaneSourceJitSafety::test_rectilinear_grid_edges_np_concrete_under_jit_tracing — uses jax.make_jaxpr to confirm edges_np() values remain concrete (not abstract tracers) when the grid is part of a traced pytree

### DIFF
--- a/src/fdtdx/core/grid.py
+++ b/src/fdtdx/core/grid.py
@@ -474,7 +474,7 @@ def calculate_time_offset_yee(
     effective_index: jax.Array | float | None = None,
     e_polarization: jax.Array | None = None,
     h_polarization: jax.Array | None = None,
-    coordinate_edges: tuple[np.ndarray | jax.Array, np.ndarray | jax.Array, np.ndarray | jax.Array] | None = None,
+    coordinate_edges: tuple[np.ndarray, np.ndarray, np.ndarray] | None = None,
     center_physical: jax.Array | None = None,
 ) -> tuple[jax.Array, jax.Array]:
     if inv_permittivities.ndim == 4:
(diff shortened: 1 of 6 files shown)


ymahlau:
Hi @teevee112, thanks for the quick fix. However, I don't like the approach of storing the grid coordinates as numpy arrays. Having them as jax arrays would have many advantages, for example one could compute many simulations on different grids (but the same number of grid points) without recompilation. 

I think a better approach would be to keep the grid as jax arrays, but only convert the grid to numpy arrays for the tidy3d call inside the external callback to prevent tracer arrays

Issue #330, opened 2026-05-11
mode overlap not working with multiple frequencies


ymahlau:
Currently, the mode overlap detector only selects the first frequency for computing a mode. Instead, it should compute the mode for all frequencies.

PR #332, opened 2026-05-12
multi-frequency mode solving with continuity tracking in ModeOverlapDetector

teevee112:
WIP

Addresses https://github.com/ymahlau/fdtdx/issues/330

### Motivation

ModeOverlapDetector previously only accepted a single WaveCharacter, requiring one FDTD run per wavelength for broadband S-parameter sweeps. This PR adds multi-frequency support so all wavelengths are solved and accumulated in a single run. Along the way, it fixes two independent correctness bugs found during review: a wrong normalization convention in bidirectional_mode_overlap and a silent area swap in RectilinearGrid.yee_face_areas for y-propagating modes on non-uniform grids.

### Implementation choices

TODO

### DIFF
--- a/src/fdtdx/__init__.py
+++ b/src/fdtdx/__init__.py
@@ -27,7 +27,6 @@ from fdtdx.core.jax.pytrees import (
 )
 from fdtdx.core.physics.losses import metric_efficiency
 from fdtdx.core.physics.metrics import (
-    bidirectional_mode_overlap,
     compute_energy,
     compute_poynting_flux,
     normalize_by_energy,
@@ -223,7 +222,6 @@ __all__ = [
     "WaveCharacter",
     "apply_params",
     "autoinit",
-    "bidirectional_mode_overlap",
     "boundary_objects_from_config",
     "calculate_sparam",
     "calculate_sparams",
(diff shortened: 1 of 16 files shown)


ymahlau:
Hi @teevee112, some comments without looking at the code: 
- One important thing, which i strongly disagree: apply params needs to be jittable. This is extremely important for inverse design, because otherwise gradient computation does not work.
- Regarding the tidy3d _dot_numpy function: Did you copy that from the tidy3d repo? If so, we cannot include it because the license of tidy3d forbids using their code in other repositories. It is only okay to include a dependency, but not include or change code directly.
- Please explain the proximity tracking in more detail. I don't understand why a discontinuity might happen, what effect this has and what your implementation changes to handle that.

ymahlau:
@teevee112 as a general comment: please split your PR into smaller parts. Supporting multi-frequency for all sources and detectors should be a simple change with <100 lines of code. In this PR the changes are muddled with other things (bugs, continuity, ...) which makes review extremely difficult. For example, in your recent big PR for re-applying non-uniform grids other changes were silently reverted (see #357) without anyone noticing. 

I will close this PR now because I do not see a way to properly give this a review. I am very grateful for your work and don't want to discourage you to make further changes on fdtdx, but it does need to happen with a little more strucure. 

Feel free to open several PRs for the changes addressed here. If you open one small PR for each of the bugs, one for the multi-frequency support I would be very happy to review and give detailed feedback. This project lives from contribution like yours!

PR #336, opened 2026-05-12
automatically detect and exclude large objects from plot_setup

### LINKED ISSUE #333: automatically exclude large objects from plot_setup slices
In plot_setup, when plotting the current setup from three different sides, there usually are some detectors that span an entire plane in two axes. They can be manually excluded, which is lengthy code for a user. instead, it would be better to automatically detect which objects take up 100% of the area when looking from a specific view and excluding them automatically.

renaissancenerd:
resolves #333

## Summary by CodeRabbit

* **Improvements**
  * Plotting now detects and automatically excludes objects that nearly cover a viewing plane to reduce visual clutter; this behavior is enabled by default and can be toggled in plotting options.
  * Automatic exclusion is applied consistently across all viewing planes and respects current object filtering.

* **Documentation**
  * Updated parameter descriptions to explain the automatic exclusion behavior and how to enable or disable it.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ymahlau/fdtdx/pull/336?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


renaissancenerd:
@ymahlau ready for merge

ymahlau:
Looks good to me!

PR #343, opened 2026-05-13
changing partial_real_position to centre

### LINKED ISSUE #326: placement via partial-real-position
Have not investigated in detail, but it seems to me that when an object is placed via partial-real-position at position (0, 0, 0) (e.g. ExtrudedPolygon), then it is placed in the lower left corner of the simulation volume, not the center (where the origin should be). This is a bug and needs investigation.

renaissancenerd:
resolves #326 

## Summary by CodeRabbit

* **Bug Fixes**
  * Real-space coordinates are now interpreted relative to the simulation volume center instead of the lower corner, providing more intuitive position specification.
  * Enhanced error messages for constraint resolution with explicit bounds information.

* **Tests**
  * Updated tests to reflect center-relative coordinate semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ymahlau/fdtdx/pull/343)

### DIFF
--- a/src/fdtdx/fdtd/initialization.py
+++ b/src/fdtdx/fdtd/initialization.py
@@ -725,20 +725,26 @@ def resolve_object_constraints(
     return resolved_slices, errors
 
 
-def _center_to_bounds(real_pos: float, resolution: float, size: int) -> tuple[int, int]:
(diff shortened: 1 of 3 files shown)


renaissancenerd:
@ymahlau ready for review

ymahlau:
Looks good to me!

PR #347, opened 2026-05-19
Fix: guard setup functions against JIT tracing + remove illegal float() cast in PML

Xenosdu78:
This is a small PR that doesn't change any already functional behavior.

In the setup functions, place_objects and its callees allocate device memory and resolve concrete grid geometry. Both are incompatible with JAX JIT tracing, but the errors produced (ConcretizationTypeError, make_array_from_single_device_arrays with tracers) were cryptic and gave no guidance.

This PR does the following:

-   Adds check_not_tracing() utility in a guards.py file
-   Calls it at the entry point of place_objects
-   Removes float() cast on sigma_end_calculated in PerfectlyMatchedLayer.place_on_grid, which caused a separate                    
    ConcretizationTypeError for the same root reason
-   Fixes minor static typing errors with None initializations and asserts

Previous error message:
ValueError: jax.make_array_from_single_device_arrays requires a list of concrete arrays as input, but got types {<class 'jax._src.interpreters.partial_eval.DynamicJaxprTracer'>}

New error message:
TypeError: fdtdx.place_objects is a setup function and cannot be called inside jax.jit(). It allocates device arrays and resolves concrete grid geometry, both of which require eager execution.

Move all fdtdx setup calls outside any JIT boundary and only JIT-compile the simulation time-stepping loop:

Correct
objects, arrays, config, _ = fdtdx.place_objects(...)
result = jax.jit(run_simulation)(arrays, ...)

Wrong
@jax.jit
def run():
objects, arrays, config, _ = fdtdx.place_objects(...) # ERROR
return run_simulation(arrays, ...)

### DIFF
--- a/src/fdtdx/core/jax/guards.py
+++ b/src/fdtdx/core/jax/guards.py
@@ -1,0 +1,32 @@
+import jax
+import jax.core
+import jax.numpy as jnp
+
+
+def check_not_tracing(fn_name: str) -> None:
+    """Raise a clear error if called inside a JAX JIT trace.
+
+    Setup functions allocate concrete device memory and resolve static grid
+    geometry. Both require eager execution and are incompatible with JIT
+    tracing. This guard surfaces that constraint with an actionable message
+    instead of a cryptic internal JAX error.
+
+    Works by checking whether a freshly created jnp array is a Tracer —
+    which it is under JIT, and is not in eager mode.
+    """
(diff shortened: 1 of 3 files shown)


ymahlau:
Hi @Xenosdu78, thanks for implementing this and sorry that it took so long for me to come back to this!

Noe-LV:
Hi @ymahlau! No problem, glad to see you're back.

PR #352, opened 2026-06-04
all init exported to docs and test

### LINKED ISSUE #297: docs: all init exported should be in docs
Currently, not all functions from init are displayed in the docs. change that. add a test that init exported functions need to be in docs/api list.

renaissancenerd:
resolves #297 

## Summary by CodeRabbit

* **Documentation**
  * Expanded the public API reference to include many additional symbols (boundaries, symmetry/grid types, conductors, PML, s‑parameter and simulation‑setup helpers, plotting/export utilities, uniform grid and symmetry entries) and adjusted a color reference; also hardened Sphinx config to preserve defaults and provide a safe fallback for problematic object descriptions.

* **Tests**
  * Added automated checks that the API docs exactly match exported public symbols, reporting missing, extra, or duplicate entries.


renaissancenerd:
@ymahlau the check is failing due to sphinx-autobuild of the docs. I did the following cmd `sh docs/scripts/sync_notebooks.sh && uv run sphinx-autobuild -W --keep-going docs/source/ docs/build/` but it does not work. Can you take a look ?

ymahlau:
@renaissancenerd, it seems like there are errors in the docstrings of the new objects that you added. Warnings are also treated as errors here. Add the objects one by one into the API list and see when it fails. Then fix the corresponding docstrings.

renaissancenerd:
@ymahlau ready for review

src/fdtdx/__init__.py
```diff
@@ -134,9 +133,21 @@
...
 
+#: Type alias for detector state: maps detector output names to JAX arrays.
+DetectorState = DetectorState
+
+#: Type alias for parameter dictionaries. Maps parameter names to JAX arrays or
+#: nested dicts of JAX arrays.
+ParameterContainer = ParameterContainer
+
+#: Type alias for simulation state: a tuple of (time_step, ArrayContainer).
+SimulationState = SimulationState
```
ymahlau, line 147: This is a bit odd, why do you redefine this here with the same name?

docs/source/07_api.rst
```diff
@@ -52,28 +70,37 @@ API
...
+    fdtdx.plot_field_slice_component
+    fdtdx.plot_material
+    fdtdx.plot_material_from_side
     fdtdx.plot_setup
     fdtdx.plot_setup_from_side
-    fdtdx.plot_setup_from_side
-    fdtdx.plot_material_from_side
-    fdtdx.plot_material
-    fdtdx.plot_field_slice_component
-    fdtdx.plot_field_slice
```
ymahlau, line 67: why are these functions removed? if they are not in the __init__ export please add them

docs/source/conf.py
```diff
@@ -91,3 +93,27 @@
     }
 }
 
+def _patch_pytreeclass() -> None:
```
ymahlau: why is this included here? please remove

docs/source/conf.py
```diff
@@ -35,7 +35,9 @@
 ]
 
 templates_path = ['_templates']
-exclude_patterns = []
+exclude_patterns = [
+"api/fdtdx.constants.rst"
```
ymahlau: please fix rather than exclude if possible

tests/docs/test_api_rst.py
```diff
@@ -0,0 +1,109 @@
...
+# Path to the RST file.
+# This test lives at tests/docs/, so go up 3 levels to reach the repo root.
+# ---------------------------------------------------------------------------
+REPO_ROOT = Path(__file__).parent.parent.parent  # tests/docs/ -> tests/ -> repo root
+API_RST = REPO_ROOT / "docs" / "source" / "07_api.rst"
+
+# Submodules exported from fdtdx (e.g. fdtdx.constants) cannot be documented
+# by autosummary cleanly — they cause docutils errors from the built-in dict
+# docstring.  Exclude them from the RST and from this check.
+_SUBMODULE_EXCLUDES: frozenset[str] = frozenset({"constants"})
```
ymahlau: again, please fix and dont exclude

docs/source/07_api.rst
renaissancenerd: they are not removed. They are still there but the uv pre-commit moved it elsewhere in the list.

src/fdtdx/__init__.py
renaissancenerd: it helped with the dict.__doc__ error and ModuleAnalyzer

renaissancenerd:
@ymahlau ready for merge

renaissancenerd:
@ymahlau i have been able to remove the wavelength dependency on the module but for the type aliases, no matter what I do, the warnings always pop up and it seems quite complicated to find the real culprit.

ymahlau:
All right, let's merge it then! Many thanks!

PR #342, opened 2026-05-13
enhancement: fixed the plot functions origins.

Xenosdu78:
Changed the following plot functions origins:

- plot_material_from_side
- plot_field_slice_component
- plot_setup_from_side

They now take the center of the grid as origin, instead of the lower left corner. The axes now have coordinates [-L/2, L/2] instead of the previous [0, L].

This choice matches the objects placement system and avoids possible confusions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Field slice plots now derive displayed bounds from transposed data so plotted orientation and coordinates align with the underlying array.
  * Material visualizations use symmetric spatial extents centered at the origin for consistent axis ranges across Z/Y/X views.
  * Plot setup now uses a centered coordinate system instead of a 0-based origin to improve spatial positioning of objects and axes.

* **Tests**
  * Integration tests updated to validate plotted extent spans (max–min) on each axis rather than only max coordinates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ymahlau/fdtdx/pull/342)

PR #344, opened 2026-05-16
Re-apply rectilinear grid support (#312) with JAX-native edge coordinates and test fixes

teevee112:
This PR re-applies #312 (which was reverted in #339), with two additional fixes on top:

1. Addressing https://github.com/ymahlau/fdtdx/issues/335 and https://github.com/ymahlau/fdtdx/issues/329: JAX-native edge coordinates — the follow-up PR #331 fixed TracerArrayConversionError by storing grid edges as Python tuples (frozen pytree aux data)
so they stay concrete under jax.jit. This is the wrong approach: grids with the same shape but different spacings would force
recompilation. Instead, this PR keeps edges as JAX arrays and passes them as explicit jax.pure_callback arguments to the tidy3d mode solver —
np.asarray() only happens inside the callback where values are always concrete, so no workaround is needed.
2. Test config fixes — unit tests that used sub-µm RectilinearGrid spacings with time=1e-8 s were generating millions of time steps, causing slow
place_on_grid calls due to O(time_steps) Python loops in OnOffSwitch. Fixed time= to 1e-13 s in the affected tests. Also updated all files still using
the removed resolution= kwarg on SimulationConfig.

Tests ran in 6 m 44 s (https://github.com/ymahlau/fdtdx/actions/runs/25975400480/job/76354551181?pr=344)

## Summary by CodeRabbit

* **New Features**
  * Support for non-uniform (rectilinear) simulation grids and VTR export for rectilinear snapshots.
  * Mode solver and sources accept explicit transverse coordinates for non-uniform meshes.

* **Bug Fixes / Improvements**
  * Plots, detectors, PML, and diagnostics now respect physical grid coordinates and use correct area/volume weighting.
  * Examples, docs, and tests updated to the new grid API.

* **Chores**
  * Grid configuration migrated from scalar resolution to explicit grid objects.

PR #357, opened 2026-06-10
Reapply plot functions origin fix & enhance GDSLayerObject

Noe-LV:
This PR reapplies a plot functions fix #342 that has been reverted by rectilinear grid support #344.

  It changes the following functions, using the new grid system: 
  - plot_material_from_side
  - plot_setup_from_side

  They now take again the center of the grid as origin, instead of the lower left corner. The axes now have coordinates [-L/2, L/2] instead of the previous [0, L].

In the current code, the color field of GDSLayerSpec is not accessed. Additionally, this PR adds a color field for GDSLayerObject and matches it to that of GDSLayerSpec. That helps controlling the setup of a simulation visually via plot_set_up.
A small sanity check of the layers attribute of GDSLayerObject is added.

## Summary by CodeRabbit

* **New Features**
  * Optional color customization for GDS layers (default applied).

* **Bug Fixes**
  * Error raised for empty layer configurations.
  * Subplot creation now enforces that axes are present.

* **Improvements**
  * Plot coordinates and slice extents are centered around the domain midpoint.

* **Tests**
  * Plot tests updated to expect centered axis limits.


Noe-LV:
Actually, this representation of the grid seems to be inconsistent with the one used for example in gds_layer_stack.py:

```
def _gds_to_sim(gds_val: float, gds_center_val: float, vol_size: float) -> float:

    """Convert a GDS coordinate (metres) to a simulation real-space coordinate (metres).

    Simulation real-space coordinates are measured from the lower-left corner of the
    simulation volume (grid index 0).
    """
    return gds_val - gds_center_val + vol_size / 2

```    
The choice of putting the coordinates origin at the center of the simulation volume or at its lower-left corner should be consistent across the entire library. Maybe we should raise a separate issue for this?

ymahlau:
Hi @Noe-LV, thanks for re-adding this PR and sorry about dropping your earlier PR. This must have slipped through the big PR. 

I agree that we should add a seperate issue for the representation. See #358. Thanks for bringing this up!

Noe-LV:
Hi @ymahlau, no problem, I understand that it slipped during the big change!
Thanks for raising up the issue.

Issue #272, opened 2026-04-08
Far-Field transformation utilities


ymahlau:
Just as the name suggests. Inspiration might be taken from chromatix (https://github.com/chromatix-team/chromatix/).

PR #375, opened 2026-06-25
Feat: add field projection detectors

Zhenhao-Xie:
## Summary

Adds three near-to-far field projection detectors:

- `FieldProjectionAngleDetector`
- `FieldProjectionCartesianDetector`
- `FieldProjectionKSpaceDetector`

These detectors project recorded frequency-domain E/H phasors from planar surfaces or box surfaces to angular, Cartesian observation-plane, or k-space direction grids. The implementation covers the main Tidy3D-style projection features, including planar and box projections.

## Tests

Added:

- Unit tests for angle, Cartesian, and k-space field projection detectors.
- JSON serialization tests for the new detector types.
- API documentation entries for the new public detector classes.

Passed:

- `uv run pre-commit run -a --show-diff-on-failure`
- `sh docs/scripts/sync_notebooks.sh && uv run sphinx-build -W --keep-going docs/source/ docs/build/`
- `uv sync --locked`
- Full local test coverage, including the `gdsfactory` checks:
  - `2704 passed`
  - `1 xfailed`

## Summary by CodeRabbit

* **New Features**
  * Added new public field-projection detectors with angle, Cartesian, and k-space projection outputs.
  * Exposed the new detectors in the package’s public API.
* **Documentation**
  * Updated Sphinx API autosummary to include the new detector classes.
* **Bug Fixes**
  * Updated JSON setup validation so the new detector types serialize and deserialize correctly.
  * Added stricter isotropic material-property handling and validation.
* **Tests**
  * Added unit and integration tests for projection behavior, JAX compatibility, JSON round-trips, and material validation.


ymahlau:
Hi @Zhenhao-Xie, thanks again for another PR! Some general comments after skimming through the code:
- please try to reduce the size of the code wherever possible. A 4000 line PR is easy to generate with AI, but hard to review and maintain long term. There is some inherent complexity, but still I think that this is a little excessive.
- Try to reduce the runtime of the tests. Your new tests alone take a few minutes to run in CI and we try to keep unit tests fast. I like that you tested a lot though, so don't remove any important tests
- Please don't only document user-facing code, but also document design decisions and the math/methods/algorithms used. Currently all the private functions are not documented, which makes it difficult to infer what is happening and why. This also helps maintainability later.

ymahlau:
One other point: please put the general jax helpers into the the core module and try to re-use the functions that are already available there. I believe there is some overlap.

Zhenhao-Xie:
@ymahlau Thanks for the suggestions. I have tried to address the maintainability and CI-runtime concerns in the latest updates.

The source implementation has been reduced by about 400 lines while keeping the detector functionality unchanged. The remaining added lines are mostly internal documentation/comments explaining the projection math.

I also refactored the shared projection/JAX logic into the core physics module. The projection-specific tests were compacted and their runtime is now roughly one third of the previous version.

The feature is still relatively large because it adds three projection detectors and supports planar/box projection, multiple coordinate systems, JAX/JIT/grad-compatible paths, and finite-distance projection. I think further substantial reduction would likely require dropping coverage or functionality. If JIT/backpropagation behavior does not need to be tested explicitly, I can reduce the test runtime further.

src/fdtdx/core/physics/field_projection.py
```diff
@@ -0,0 +1,274 @@
...
+    sin_phi: _ProjectionArray,
+    cos_phi: _ProjectionArray,
+) -> tuple[jax.Array, jax.Array, jax.Array]:
+    """Build global spherical basis vectors from precomputed sine and cosine arrays.
+
+    Both tensor-product angle grids and paired angle arrays share this helper. The inputs may already contain
+    broadcast dimensions; the output component axis is always first.
+    """
+    zeros = jnp.zeros_like(sin_theta * cos_phi, dtype=float)
+    radial = jnp.stack(
```
ymahlau: Can you put this on one line just for better formatting?

src/fdtdx/core/physics/field_projection.py
```diff
@@ -0,0 +1,274 @@
...
+``I(r_hat) = int_S J(r_prime) exp(-i k r_hat . r_prime) dS``. The exact finite-distance path evaluates the
+homogeneous-medium dyadic Green-function fields from equivalent electric and magnetic surface currents. Both paths
+use physical detector coordinates and trapezoidal weights, which is important for rectilinear non-uniform grids.
+"""
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+_ProjectionArray = jax.Array | np.ndarray
```
ymahlau: I think a bit unnecessary to introduce private types for this. Just repeat the "jax.Array | np.ndarray" everywhere where it is used

src/fdtdx/objects/detectors/field_projection.py
```diff
@@ -0,0 +1,1559 @@
...
+    subsample_indices,
+    trapezoidal_weights_1d,
+)
+from fdtdx.materials import Material
+from fdtdx.objects.detectors.detector import DetectorState
+from fdtdx.objects.detectors.phasor import PhasorDetector
+from fdtdx.typing import SliceTuple3D
+
+ProjectionSurface = Literal["x-", "x+", "y-", "y+", "z-", "z+"]
+ProjectionArray = jax.Array | np.ndarray
```
ymahlau: Again, ProjectionArray and ProjectionScalar should be removed. ProjectionSurface is a good type, but it should live in `typing.py`

src/fdtdx/objects/detectors/field_projection.py
```diff
@@ -0,0 +1,1559 @@
...
+    return _SURFACE_NAMES_BY_AXIS_DIRECTION[(axis, direction)]
+
+
+def _surface_state_key(surface: ProjectionSurface) -> str:
+    """Return the state-dictionary key used to store a box-face phasor."""
+    return f"phasor_{surface[0]}_{_SURFACE_KEY_SUFFIX[surface[1]]}"
+
+
+def _contains_bool(values: Any) -> bool:
+    """Return whether a scalar or nested container contains boolean values.
```
ymahlau: belongs to core module

src/fdtdx/objects/detectors/field_projection.py
```diff
@@ -0,0 +1,1559 @@
...
+
+def _finite_scalar_or_1d_array(name: str, value: Any, expected_size: int) -> float | np.ndarray:
+    """Validate a scalar or one value per wave character for projection-medium settings."""
+    message = f"{name} must be a finite numeric scalar or contain {expected_size} values."
+    array = _real_numeric_array(name, value, message, keep_jax=False)
+    if array.ndim == 0:
+        return float(array)
+    if array.shape != (expected_size,):
+        raise ValueError(message)
+    return cast(np.ndarray, array)
```
ymahlau: everything up to here belongs in core module

src/fdtdx/core/physics/field_projection.py
```diff
@@ -0,0 +1,274 @@
...
+def edge_window_1d(points: _ProjectionArray, window_size: float) -> jax.Array:
+    """Return a Gaussian finite-aperture taper that suppresses only the two edges.
+
+    ``window_size`` is the fractional width of the tapered region over both edges. The central region is left at
+    unit weight, which keeps the main aperture amplitude unchanged while reducing edge-diffraction ringing.
+    """
+    points = jnp.asarray(points, dtype=float)
+    if window_size <= 0:
+        return jnp.ones(points.shape, dtype=float)
+    window_factor = 15.0
```
ymahlau: where does this magic number come from?

ymahlau:
Some other comments:
- duplication between _surface_currents_geometry (lines 863-913) and the subsample/weight block at the top of _project_surface_far_field_grid (lines 942-951). Both recompute subsample_indices, slice the fields, and build u_weights/v_weights/weights with identical code.
- The thin one-line accessor wrappers _projection_refractive_index, _projection_impedance, _projection_wavenumber (lines 819-832) each just unpack one element of _projection_parameters. Callers could unpack directly; these add surface area without much value.

Also let Claude find some duplications:
1. Axis-0 cross product — clearest duplication. _cross_axis0 in core/physics/field_projection.py:179 reimplements a component-axis-first cross product using moveaxis round-trips.
But core/physics/metrics.py already does exactly this, more cheaply, via JAX's native axis arguments: jnp.cross(..., axisa=0, axisb=0, axisc=0) (lines 111 and 138). The moveaxis wrapper is functionally identical to passing axisa=axisb=axisc=0 and is strictly slower. This should either call a shared one-liner or just inline the axis*=0 form. The two files even live in the same core/physics package.
2. Isotropic-material extraction — duplicates Material's own logic. _isotropic_material_property_scalar (detector file, lines 234-255) hardcodes _MATERIAL_DIAGONAL_INDICES = (0, 4, 8) and _MATERIAL_OFF_DIAGONAL_INDICES = (1, 2, 3, 5, 6, 7), then manually checks off-diagonals are zero and diagonals are equal. The Material class already owns this exact knowledge: materials.py:381 has _is_property_isotropic(prop) which does the same flat-9 off-diagonal/diagonal check with math.isclose, and Material.is_all_isotropic() (line 209) wraps it across all four properties. The PR is re-deriving the 3×3-stored-as-9 representation instead of asking the material whether it's isotropic and pulling the scalar. This is the overlap the maintainer most likely meant by "I believe there is some overlap" — it's not just style, it's re-encoding the material layout in a second place, so a future change to that layout breaks the projection detector silently.
3. expand_to_3x3 already handles the 9→tensor reshape. core/misc.py:515 (expand_to_3x3) is the canonical place that interprets the (1/3/9, ...) material storage, including the isotropic and full-tensor cases. The PR's manual raw_array.shape != (9,) handling overlaps conceptually with this. Less critical than (2) since the projection wants a scalar rather than a 3×3, but it reinforces that material-shape interpretation belongs in one module.

src/fdtdx/objects/detectors/field_projection.py
```diff
@@ -0,0 +1,1559 @@
...
+        results = [
+            self._project_observation_angle_grid(
+                state,
+                theta,
+                phi,
+                wave_character_index=wave_character_index,
+            )
+            for wave_character_index in range(len(self.wave_characters))
+        ]
+        return self._stack_frequency_results(
```
ymahlau: Another general comment: Please keep the code readable. Instructions are often scattered across multiple lines. Here for example, one could define the second return value dictionary in one line and the have the return statement in the second line. Scattering everythin over five lines makes it less readable, especially since there are no keyword arguments used.

ymahlau:
Thanks for the update @Zhenhao-Xie! I think the balance between test coverage and speed is good now. 

I did a more detailed review of the code and left some comments above. There is nothing major, just a bit of code styling and maintainability.

After adressing these comments, the PR should be ready to be merged!

ymahlau:
relates to #272

Zhenhao-Xie:
@ymahlau Thanks for the careful review. 

I have addressed the remaining maintainability comments in the latest commits. The general helper logic has been moved into the appropriate core modules.  I also cleaned up the other points you mentioned: removed duplicated projection code, simplified thin wrappers, reused existing core functionality, and adjusted the formatting/readability issues. 

I think this should now be ready for merge.

ymahlau:
Looks good to me! Many thanks @Zhenhao-Xie!

Issue #368, opened 2026-06-22
Subpixel Averaging


bruxillensis:
Is anyone working on this? Would it be alright if I did? Any effect on other work/issues ongoing right now?

l-berg:
Hi. In general, any work on subpixel averaging would be highly appreciated.

I am working on some minor adjacent details: Currently, FDTDX uses linear interpolation of the inverse permittivities to interpolate between materials (for continuous devices and mulit-material objects that use a `voxel_mask`). I am changing this to use permittivities instead of inverse permittivities, and also adding etching support for devices (i.e., support devices that do not fully define their `grid_shape`, but leave some of the existing materials intact).

These changes should hopefully come later this week, are only minor in scope and do not affect any interfaces.

l-berg:
Not sure whether this is relevant, but I also created a smooth version of the Polygon. It uses shapely to compute the fraction of how much each pixel is covered by the polygon. Since shapely scales poorly, I use the matplotlib Path test for most pixels, and shapely only for border pixels.

```
import shapely
from matplotlib.path import Path
from scipy.ndimage import binary_dilation, binary_erosion
def polygon_to_mask_hybrid(
    boundary: tuple[float, float, float, float],
    resolution: float,
    polygon_vertices: np.ndarray,
    safe_margin: int = 2
) -> np.ndarray:
    """
    Generate a 2D fractional mask, highly optimized for large grids.
    Mixes fast binary point-in-polygon with exact Shapely boundary intersections.
    """
    assert polygon_vertices.ndim == 2
    assert polygon_vertices.shape[1] == 2
    min_x, min_y, max_x, max_y = boundary

    # 1. Create coordinate arrays
    x_coords = np.arange(min_x, max_x + 0.5 * resolution, resolution)
    y_coords = np.arange(min_y, max_y + 0.5 * resolution, resolution)
    X, Y = np.meshgrid(x_coords, y_coords, indexing="ij")

    # 2. Fast Binary Masking (Core Interior/Exterior)
    points = np.column_stack((X.ravel(), Y.ravel()))
    polygon_path = Path(polygon_vertices)
    inside_polygon = polygon_path.contains_points(points, radius=1e-3*resolution)
    binary_mask = inside_polygon.reshape(X.shape)

    # 3. Identify the Boundary Band using Scipy Morphology
    # Dilation pushes the boundary out, erosion pulls it in. 
    # The difference between them is our "uncertainty band" where edges lie.
[...]

PR #378, opened 2026-06-29
Spacing-weight the off-diagonal anisotropic average on non-uniform grids

Jinge-Wang:
• `avg_anisotropic_E/H_component` used to colocate the off-diagonal field and curl components of the full-tensor anisotropic update with a plain four-point mean. On a non-uniform grid, the unweighted mean is only first-order accurate. Weighting the center-to-edge half-step of each average by the local cell widths restores second order.

• `get_anisotropic_averaging_widths()` calculates the weights from the resolved grid. It is threaded into the forward and reverse E/H updates. They depend only on the static grid, so under JIT they constant-fold and add no per-step cost.

• Tests:
Uniform grids reproduce the unweighted mean exactly; a refinement study on a graded mesh shows the weighted average converging at second order versus first order for the unweighted mean.

• Scope: 
This is the off-diagonal field/curl averaging in the full-tensor anisotropic update (fdtd/misc.py), distinct from subpixel averaging (#368) and the material permittivity interpolation around #376, which blend materials at interfaces during setup. This PR doesn't touch material smoothing.

## Summary by CodeRabbit

* **New Features**
  * Added support for anisotropic averaging on non-uniform grids, improving results when grid spacing varies.
  * Updated field update paths to use spacing-aware averaging automatically where applicable.

* **Bug Fixes**
  * Kept existing behavior unchanged for uniform grids, preserving the same averaging results as before.
  * Improved numerical accuracy on graded meshes with better convergence behavior.

* **Tests**
  * Added coverage for uniform-grid parity and weighted averaging accuracy on strongly graded meshes.

### DIFF
--- a/src/fdtdx/fdtd/misc.py
+++ b/src/fdtdx/fdtd/misc.py
@@ -133,51 +133,81 @@ def avg_anisotropic_E_component(
     field: jax.Array,
     component: int,
     location: int,
+    aniso_widths: tuple[jax.Array, jax.Array, jax.Array] | None = None,
 ) -> jax.Array:
-    """Averages an E field component at a given location.
+    """Averages an E field component onto another component's Yee location.
+
+    On a uniform grid this is the four-point mean of the staggered samples. On a non-uniform
+    grid the center-to-edge half-step along the component axis is weighted by the local cell
+    widths; the location axis is already edge-aligned and keeps an unweighted midpoint.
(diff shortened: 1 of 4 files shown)


ymahlau:
Amazing PR @Jinge-Wang! I really have no comments, everything looks good to me. Many thanks!

PR #384, opened 2026-07-01
Refactor PML: attach constants to obj, and update contiguous slices

### LINKED ISSUE #315: Refactor: PML Fields
Currently the PML fields are very ressource hungry. We should refactor them back to only be defined for the regions where the PML is actually specified. However, we should do it in such a way that all pmls are calculated in parallel. In the past it was a problem that doing a `for`-loop over pmls was not parallelized by jax if every run in the loop updates the arraycontainer.

l-berg:
Resolves #315.

This pull request further refactors the implementation of the PML boundaries to drastically minimize memory consumption and improve execution speeds for large scale single device simulations.

The architecture shifts from a sparse flat shell (via index-list `pml_indices`) to updates of continguous slices. Here are the main changes:
* **Localized Coeffs & Memory:** The CPML recurrence coefficients ($a, b, \kappa^{-1}$) are now cleanly initialized and cached locally within each PML object instance using its precise spatial slice (`grid_slice`).
* **Per-Object Loop Execution:** The global sparse indices have been completely stripped out. Instead, `curl_E` and `curl_H` iterate over `objects.pml_objects` to execute `step_cpml()`, modifying only the relevant multi-dimensional subarrays directly via local slices.
* **Granular Auxiliary Fields:** The global `(6, M)` array fields for `psi_E` and `psi_H` have been replaced with a dictionary map (`PmlAuxField`), storing explicit, isolated pairs of auxiliary tensors bound strictly to each layer's identifier and local grid shape.

By eliminating the dense 3D arrays during initialization and the index-gathering, memory requirements are cut down dramatically from 17 GB to 4.7 GB in one of my experiments with two volume shapes of (520, 320, 120) each. These reductions are most notable with large simulation volumes.

Additionally, using known contiguous slices rather than gathering dynamically with `.at[:, ix, iy, iz].add` results in a runtime speedup. In my experiments we go from simulation times of 470s to 320s (versus 260s without PML). In this case, PML runtime overhead was reduced by ~70%.

However, there may be tradeoffs: I did not run performance tests for small simulations, and I was not able to test the multi-GPU case. Especially the last one might suffer because of the overhead introduced by more sequential communication.

## Summary by CodeRabbit

* **New Features**
  * Improved perfectly matched layer (PML) handling: boundary corrections are now applied locally during electric and magnetic curl calculations.
  * PML auxiliary state is tracked per PML region, simplifying initialization and updates for PML-enabled simulations.

* **Bug Fixes**
  * Updated forward and reverse FDTD update paths to consistently use the new PML correction workflow.
  * Adjusted simulation state and wrapper behavior so PML state is correctly represented during gradient/time-reversal runs.

* **Tests**
  * Updated/extended PML and curl-related test coverage to match the new PML auxiliary-field behavior.


ymahlau:
Looks good to me @l-berg

PR #363, opened 2026-06-16
Quasi-uniform grid support and coordinates origin at center

Noe-LV:
This PR creates a new grid policy called QuasiUniformGrid that lets the user set a constant mesh resolution in each space direction. It takes as arguments dx, dy, and dz that correspond to these constant spacings.

Importantly, this PR changes the order of operations during constraints resolution. All grid policies are first resolved as Rectilinear grids, using only the volume and grid policy specifications. Then placement constraints are resolved with a Rectilinear grid. This is consistent with the choice of keeping apart user-friendly grid policy classes such as UniformGrid, QuasiUniformGrid, possibly AutoGrid as in tidy3d, and an internal RectilinearGrid representation for the algorithms.

It also partly adresses #358 and tries to keep real-space coordinates consistent  with the representation choice of origin-at-the-center coordinates. However, RectilinearGrid coordinates are still computed as positive integers (e.g. for a cubic volume with length of 100µm and 1µm resolution, real coordinates are (-50e-6, 50e-6), but RectilinearGrid coordinates are (0, 100)). This needs to be updated.

Tests were updated for the most part. Some still use origin-at-lower-corner coordinates.

## Summary by CodeRabbit

* **New Features**
  * Added `QuasiUniformGrid` support with independent per-axis spacing, validation, and CFL-based timestep computation.
* **Improvements**
  * Updated the simulation domain to a center-based coordinate convention, improving resolved-grid snapping and geometry calculations.
  * Enhanced object placement and constraint solving by resolving and using the final grid geometry earlier.
  * Adjusted GDS coordinate alignment and voxelization to match the center-based convention.
* **Tests**
  * Updated unit/integration tests for the new centering, snapping, and expected coordinate/bounds behavior.

PR #386, opened 2026-07-02
Fix GDS layer masks on rectilinear grids

teevee112:
This is a small follow-up to #363, which moved grid coordinates toward a center-origin convention and added rectilinear/quasi-uniform grid support.

The GDS layer code already converts imported polygons into gds_center-relative coordinates:

local_polygon = polygon - gds_center

The uniform-grid path samples cell centers in that same center-relative frame, so it stayed consistent after #363. The rectilinear-grid path, however, was still passing absolute grid cell centers into
polygon_to_mask_at_points. That meant it could compare center-relative polygon coordinates against absolute rectilinear grid coordinates.

Example:

1. A rectilinear grid has edges from 0 nm to 1000 nm, so the object center is at 500 nm.
2. gds_center = 400 nm, meaning GDS 400 nm should map to the object center.
3. A GDS polygon from 300 nm to 500 nm becomes local coordinates from -100 nm to 100 nm.
4. Before this fix, the rectilinear path tested that local polygon against absolute cell centers around 400-600 nm.
5. After this fix, the rectilinear path subtracts the object slice center first, so those sample points become center-relative and match the polygon frame.

This keeps the intent of #363 while making the GDS rectilinear voxelization path consistent with the uniform-grid path and the documented gds_center behavior.

Regression coverage added for:

- nonzero gds_center placement on rectilinear grids
- uniform-grid vs rectilinear-grid parity
- nonuniform rectilinear cell-center sampling

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved voxel mask placement accuracy for GDS-derived shapes, including correct handling of per-slice offsets and non-uniform rectilinear grids.
  * Corrected coordinate conversion so port positions map consistently into simulation space.
  * Updated alignment behavior for `gds_center` to reflect placement relative to the object’s slice center.
* **Tests**
  * Added new unit tests for offset `gds_center` scenarios and verification against expected boolean masks on mixed-resolution rectilinear grids.

### DIFF
--- a/src/fdtdx/objects/static_material/gds_layer_stack.py
+++ b/src/fdtdx/objects/static_material/gds_layer_stack.py
@@ -156,9 +156,13 @@ class GDSLayerObject(StaticMultiMaterialObject):
             v_lower, v_upper = self.grid_slice_tuple[self.vertical_axis]
             h_edges = np.asarray(grid.edges(self.horizontal_axis))
             v_edges = np.asarray(grid.edges(self.vertical_axis))
-            # Compute cell centers relative to the origin at the center of the simulation volume
-            h_centers = 0.5 * (h_edges[h_lower:h_upper] + h_edges[h_lower + 1 : h_upper + 1])
-            v_centers = 0.5 * (v_edges[v_lower:v_upper] + v_edges[v_lower + 1 : v_upper + 1])
+            # local_polygons are in gds_center-relative space (poly - gds_center).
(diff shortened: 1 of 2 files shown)


ymahlau:
Hi @teevee112, thanks for fixing this bug! One minor thing: The class docstring at lines 70 is now still describing the wrong old behavior of placement at the simulation center, not object center. Can you fix that as well please?

Noe-LV:
Hi @teevee112 , thanks for the follow-up!

PR #389, opened 2026-07-06
Fix inconsistent mode scaling between source and detector

l-berg:
The recent PR #388 reverted the mode normalization, so that the fields produced by a mode source are suitable for checkpointing in smaller dtypes. However, this introduced an inconsistency between mode source and mode detector.

This PR makes the behaviour consistent, so that a mode detector placed on the same waveguide as a mode source measures a mode overlap of 1, by applying the same scaling to the mode overlap computation in the mode detector that is also applied in `compute_mode`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mode-overlap calculations by normalizing detector face-area weighting, making results consistent with the scaling used during mode computation.
  * Overlap outputs are now invariant to the absolute detector face area when the overlap integrand is constant.
* **Tests**
  * Updated unit tests to reflect the corrected weighting behavior and adjusted expected scaling accordingly.

### DIFF
--- a/src/fdtdx/objects/detectors/mode.py
+++ b/src/fdtdx/objects/detectors/mode.py
@@ -104,6 +104,7 @@ class ModeOverlapDetector(PhasorDetector):
         else:
             spacing = self._config.uniform_spacing()
             weights = jnp.ones(self.grid_shape, dtype=jnp.float32) * spacing * spacing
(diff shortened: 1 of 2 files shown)


ymahlau:
Thanks @l-berg

PR #391, opened 2026-07-06
enhancement: "auto" aspect plots and color map argument for detector videos

Noe-LV:
This small PR adds two arguments to detector states plotting functions:

- `aspect` that can be either `auto` or `equal`. The legacy behavior, `equal`, sets equal scales to all axes when plotting figures. It is kept as default. The other behavior, `auto` sets custom scales to all axes so that the plots fill the most space available.
- `cmap` is a matplotlib colormap. The legacy behavior is a custom red-blue colormap that is kept as "default".

Below is an example of plot obtained with `aspect="auto"` and `cmap="turbo"`.

Screenshot 2026-07-06 160557

And here below, an example with `aspect="equal"` and `cmap="default"`.

Screenshot 2026-07-06 165117

## Summary by CodeRabbit

* **New Features**
  * Detector 2D slice plots and video frames now support customizable `cmap` (default `default`) and `aspect` (`auto`/`equal`, default `equal`), applied consistently across plot panels and rendering modes.
* **Bug Fixes**
  * Improved plot robustness by validating required per-slice min/max values before rendering.
  * Aspect handling is now honored for both image and mesh-based views, and centered micrometer extents are used when coordinate edges aren’t provided.
* **Tests**
  * Updated unit tests for compatibility with the expanded plotting interfaces.

### DIFF
--- a/src/fdtdx/objects/detectors/plotting/plot2d.py
+++ b/src/fdtdx/objects/detectors/plotting/plot2d.py
@@ -71,7 +71,7 @@ def plot_2d_from_slices(
         extent1 = (
             -xy_slice.shape[0] * res_x / 2,
             xy_slice.shape[0] * res_x / 2,
-            -xy_slice.shape[0] * res_x / 2,
+            -xy_slice.shape[1] * res_x / 2,
             xy_slice.shape[1] * res_y / 2,
         )
         cax1 = ax1.imshow(
@@ -108,7 +108,7 @@ def plot_2d_from_slices(
         extent2 = (
             -xz_slice.shape[0] * res_x / 2,
             xz_slice.shape[0] * res_x / 2,
-            -xz_slice.shape[0] * res_x / 2,
+            -xz_slice.shape[1] * res_x / 2,
             xz_slice.shape[1] * res_z / 2,
         )
         cax2 = ax2.imshow(
@@ -144,8 +144,8 @@ def plot_2d_from_slices(
         extent3 = (
             -yz_slice.shape[0] * res_y / 2,
             yz_slice.shape[0] * res_y / 2,
-            -yz_slice.shape[0] * res_y / 2,
-            yz_slice.shape[0] * res_y / 2,
+            -yz_slice.shape[1] * res_y / 2,
+            yz_slice.shape[1] * res_y / 2,
         )
         cax3 = ax3.imshow(
             yz_slice.T,
--- a/src/fdtdx/objects/detectors/plotting/plot2d.py
+++ b/src/fdtdx/objects/detectors/plotting/plot2d.py
@@ -71,7 +71,7 @@ def plot_2d_from_slices(
         extent1 = (
             -xy_slice.shape[0] * res_x / 2,
             xy_slice.shape[0] * res_x / 2,
-            -xy_slice.shape[1] * res_x / 2,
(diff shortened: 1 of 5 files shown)


src/fdtdx/objects/detectors/plotting/plot2d.py
```diff
@@ -46,42 +55,46 @@ def plot_2d_from_slices(
...
+    assert minvals[0] is not None and maxvals[0] is not None
     if coordinate_edges_um is None:
         extent1 = (
-            0,
-            xy_slice.shape[0] * res_x,
-            0,
-            xy_slice.shape[1] * res_y,
+            -xy_slice.shape[0] * res_x / 2,
+            xy_slice.shape[0] * res_x / 2,
+            -xy_slice.shape[1] * res_x / 2,
```
ymahlau: This should be res_y here right?

src/fdtdx/objects/detectors/plotting/plot2d.py
```diff
@@ -90,69 +103,71 @@ def plot_2d_from_slices(
...
+    assert minvals[1] is not None and maxvals[1] is not None
     if coordinate_edges_um is None:
         extent2 = (
-            0,
-            xz_slice.shape[0] * res_x,
-            0,
-            xz_slice.shape[1] * res_z,
+            -xz_slice.shape[0] * res_x / 2,
+            xz_slice.shape[0] * res_x / 2,
+            -xz_slice.shape[1] * res_x / 2,
```
ymahlau: and this here res_z?

src/fdtdx/objects/detectors/plotting/plot2d.py
```diff
@@ -90,69 +103,71 @@ def plot_2d_from_slices(
...
     if coordinate_edges_um is None:
         extent3 = (
-            0,
-            yz_slice.shape[0] * res_y,
-            0,
-            yz_slice.shape[1] * res_z,
+            -yz_slice.shape[0] * res_y / 2,
+            yz_slice.shape[0] * res_y / 2,
+            -yz_slice.shape[1] * res_y / 2,
+            yz_slice.shape[1] * res_y / 2,
```
ymahlau: and the lower of these here res_z? I am not sure if I misunderstand the intent of the code, but I think your axes are tangled up

ymahlau:
Hi @Noe-LV, thanks for your PR! I think there are some issue with using this PR with non-uniform grids (which is a pre-existing issue, but should be fixed here). 

I also create a PR for defaults (see #396 ), but I think that is a different issue that should be addressed seperately

Noe-LV:
Hi @ymahlau, thanks for the review! I fixed the typos in the last commits, I went a bit too fast. Could you specify the pre-existing issues that can be encountered when using non-uniform grids? 

The plotting is handled by a separate `pcolormesh` from matplotlib when using non-uniform grids. 

I could obtain the plot here below using a QuasiUniformGrid with dx=50nm, dy=25nm and dz=100nm. So I guess the issues come up with grids that aren't uniform in any space direction?

Screenshot 2026-07-08 120139

ymahlau:
Thanks for fixing @Noe-LV, the pre-existing issues were basically that plotting did not respect non-uniform grids properly, but I think that may actually be resolved

PR #393, opened 2026-07-07
fix: serialize NumPy/JAX arrays in JSON export and allow roundtrip of polygon geometry

teevee112:
Problem
The JSON export/import pipeline (export_json_str / import_from_json) was broken for any simulation setup involving GDS-imported geometry (GDSLayerObject) or custom polygon geometry (ExtrudedPolygon). Two distinct bugs caused failures:

1. Array serialization crash (conversion/json.py)
GDSLayerObject.polygons and ExtrudedPolygon.vertices are stored as numpy.ndarrays. The JSON exporter reached the if obj in JAX_DTYPES type-check before handling arrays, which internally triggers an element-wise == comparison, resulting in:

TypeError: descriptor '__array_wrap__' for 'numpy.generic' objects doesn't apply to a 'numpy.ndarray' object

2. Deserialization shape validation error (polygon.py)
When a serialized ExtrudedPolygon is reloaded, the constructor is called with the auto-derived partial_real_shape values (e.g. (1e-6, 1e-6, 220e-9)). The __post_init__ guard was too strict and raised an error for any non-None shape, even when the value matched the vertex bounding box exactly:

Exception: ExtrudedPolygon test_poly: partial_real_shape for axis 0 is derived from the vertex bounding box. Do not specify it explicitly.

## Summary by CodeRabbit

* **New Features**
  * Improved JSON import/export for array data, including reliable round-tripping of NumPy and JAX arrays (with correct reconstruction of list-based and scalar payloads).

* **Bug Fixes**
  * Fixed JSON deserialization issues involving nested list payloads and boolean values.
  * Relaxed polygon shape validation so partial 3D real-shape inputs are only rejected when they conflict with the derived geometry.

* **Tests**
  * Added integration and unit coverage for JSON round-trip behavior of polygon materials and array inputs.

### DIFF
--- a/src/fdtdx/conversion/json.py
+++ b/src/fdtdx/conversion/json.py
@@ -6,6 +6,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence
 
 import jax
+import numpy as np
 
 from fdtdx.config import SimulationConfig
 from fdtdx.core.jax.pytrees import TreeClass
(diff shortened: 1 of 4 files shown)


ymahlau:
Hi @teevee112, thanks for reporting and directly fixing this bug. I agree with the coderabbit issue above that 0d numpy arrays should also be supported. I think storing shape metadata is probably the best solution for this.

Apart from that everything looks good to me!

ymahlau:
Awesome!

Issue #396, opened 2026-07-08
Colormap defaults


ymahlau:
I believe that having good default values is important for a library to be easily usable. Currently fdtdx uses a specific diverging color map for plots as defaults, which does not look good. 

I think the default should be a perceptually uniform color map for unsigned data (like energy), here I would go for "inferno".

For unsigned data, probably "bwr" is best, and we should make sure that vmin=vmax such that the zero-point is by default always white entirely. 

If someone has other opinions I am also happy to talk about that

ymahlau:
relates to #391

Noe-LV:
Hi @ymahlau, I'm writing a small PR to tackle this issue. I agree on the fact that the choice of color maps matters for usability.
Signed data indeed needs a diverging color map, while unsigned data needs a sequential color map.

For the choice of the actual default values, I'd tend to prefer "turbo" for the unsigned case. It is of course also a matter of personal preference. One of the reasons for this: "turbo" is known to make low lightness zones more distinguishable thanks to different color shades. See [this](https://research.google/blog/turbo-an-improved-rainbow-colormap-for-visualization/) article for reference. It describes quite well the different advantages of common colormaps.

Here below are some examples of plots for both types of data in fdtdx.

- An example of unsigned data: energy propagating in a waveguide escalator.

Turbo:
<img width="1539" height="770" alt="Image" src="https://github.com/user-attachments/assets/365a5db0-b9df-4baa-8ec0-cc5a21b84f4a" />
Inferno:
<img width="1529" height="768" alt="Image" src="https://github.com/user-attachments/assets/a33eff45-a687-4643-ac88-e451b6b2d522" />
As you can see the low-energy zones linked to radiation in the cladding appear more clearly using "turbo".

- Here are examples of signed data comparing "turbo" and "RdBl_r" in both cases. I think RdBlu_r looks better in this case.

<img width="1464" height="990" alt="Image" src="https://github.com/user-attachments/assets/f6a1e67e-e450-41c4-bd87-33254fd6224d" />

<img width="1464" height="990" alt="Image" src="https://github.com/user-attachments/assets/ef313921-e503-4973-99f3-989de4fa81c0" />

What's your opinion on all of that?

ymahlau:
@Noe-LV thanks for tackling this, I agree with your preferences of colormaps!

PR #397, opened 2026-07-08
enhancement: implements colormap defaults for signed and unsigned detector data (solves #396)

Noe-LV:
Solves #396.

A new boolean class variable, `_signed_data` is added to the `Detector` class and set as True by default.
Classes for unsigned data like energy and Poynting flux (please tell me if I forgot any), override the default value and set it as False.

Plotting colormaps are then assigned. Depending on its value, a red-white-blue color map (RdBl_r) is used when data is signed, and "turbo" is used if it is unsigned.
Max and minimum color values are set from the data: `vmin=0`, `vmax=data_max` in unsigned case and `vmin=-abs_max`, `vmax=abs_max` in signed case, where `abs_max = max(abs(data_min), abs(data_max))`. This way the zeroes points of the data are mapped to white, the 0 point of the colorbar.

## Summary of changes

* **New Features**
  * Detector plot and video rendering now automatically adjust colormaps and min/max scaling based on whether outputs can be signed or are strictly non-negative.
  * Field component visualizations now render with zero-centered color scaling for improved contrast.

* **Bug Fixes**
  * Improved auto min/max behavior: signed data defaults to symmetric bounds around 0 when limits are missing; unsigned data defaults to non-negative bounds.
  * Increased plotting robustness by ensuring required axes are present before drawing component panels.

### DIFF
--- a/src/fdtdx/objects/detectors/plotting/plot2d.py
+++ b/src/fdtdx/objects/detectors/plotting/plot2d.py
@@ -2,7 +2,6 @@ from typing import Literal
 
 import matplotlib.pyplot as plt
 import numpy as np
-import seaborn as sns
 from matplotlib.figure import Figure
 
 
@@ -18,6 +17,7 @@ def plot_2d_from_slices(
     plot_dpi: int | None = None,
     aspect: Literal["equal", "auto"] = "equal",
     cmap: str = "default",
+    signed_data: bool = True,
 ) -> Figure:
     """Plot orthogonal slices using either uniform spacing or rectilinear edges.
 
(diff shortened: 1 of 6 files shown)


src/fdtdx/objects/detectors/plotting/plot2d.py
```diff
@@ -34,31 +34,47 @@ def plot_2d_from_slices(
         maxvals: Per-slice color maxima.
         plot_interpolation: Interpolation used by the legacy uniform imshow path.
         plot_dpi: Figure DPI.
-        cmap: str = "default": Color map for the detector plots. "default" is a custom
-            red-blue seaborn color map.
+        cmap: str = "default": Color map for the detector plots. "default" is turbo
+            for unsigned data and bwr for signed data.
```
ymahlau: bwr is not used

src/fdtdx/objects/detectors/plotting/video.py
```diff
@@ -109,25 +112,35 @@ def generate_video_from_slices(
...
-        if minvals[a] is None:
-            min_list = list(minvals)
-            min_list[a] = slices[a].min()
-            minvals = min_list[0], min_list[1], min_list[2]
-        if maxvals[a] is None:
-            max_list = list(maxvals)
-            max_list[a] = slices[a].max()
-            maxvals = max_list[0], max_list[1], max_list[2]
+        if signed_data:
+            if minvals[a] is None or maxvals[a] is None:
```
ymahlau, line 132: This discards user input, I think if a single bound is given, the other bound should be set symmetrically automatically, rather than discarding the user input

ymahlau:
Thanks @Noe-LV, apart from the comments above the PR looks good to me!

Noe-LV:
Thanks for the review @ymahlau, if the tests work out correctly if think it's ready to merge now.

ymahlau:
@Noe-LV there is one test failing now, please fix that before we merge

Noe-LV:
Hi @ymahlau, I think it's ready to merge now!

PR #398, opened 2026-07-09
 add FDTD performance benchmark suite

teevee112:
FDTDX had no repeatable way to measure raw FDTD throughput or catch performance regressions on GPU. This adds a `tests/performance` benchmark suite inspired by https://arxiv.org/abs/2506.16665:
- Simulations:
  - **`directional_coupler`** — a GDS-imported SOI directional coupler, exercising a realistic workload: non-uniform grid, mode source/detectors, S-parameter extraction via `compute_overlap()`.
- environment capture (git SHA, JAX/device info), JIT compile/run separated so compile time and run time are measured independently, optional XLA trace + memory profiling, and a JSON results sink for tracking MCUPS/memory over time.
- --perf-max-steps N caps how many FDTD steps a benchmark runs, for fast diagnostic/dev runs instead of waiting out a full benchmark.
- Outputs generated in `results/` and gitignored.
- No changes to the functionality of FDTDX.

### Example run

Summary:

```
si_coupler_gds_cpl10:
  grid=(948, 145, 65)        (~8.9M cells)
  steps=11578
  compile=0.87s               (cached)
  run_best=259.525s
  MCUPS=398.6
  static_mem=839MB
  peak_mem=2475MB
```

Raw results: 

image

```
{
  "session_timestamp": "20260702T065244Z",
  "env": {
    "git_sha": "c075199",
    "jax_version": "0.10.1",
    "jaxlib_version": "0.10.1",
    "python_version": "3.14.4",
    "platform": "linux",
    "cpu_model": "Intel(R) Core(TM) i7-14700F",
    "cpu_count_logical": 28,
    "devices": [
      {
        "id": 0,
        "kind": "NVIDIA GeForce RTX 4060 Ti",
        "platform": "gpu",
        "memory_bytes": 6438912000
      }
    ],
    "jax_x64_enabled": false
  },
  "results": [
    {
      "name": "si_coupler_gds_cpl10",
      "grid_shape": [
        948,
        145,
        65
      ],
      "time_steps": 11578,
      "total_cell_updates": 103448272200,
      "compile_seconds": 0.8732664029994339,
      "run_seconds_best": 259.5251068769976,
      "run_seconds_median": 259.5251068769976,
      "run_seconds_worst": 259.5251068769976,
      "mcups": 398.60602869929454,
      "env": {
        "git_sha": "c075199",
        "jax_version": "0.10.1",
        "jaxlib_version": "0.10.1",
        "python_version": "3.14.4",
        "platform": "linux",
        "cpu_model": "Intel(R) Core(TM) i7-14700F",
        "cpu_count_logical": 28,
        "devices": [
          {
            "id": 0,
            "kind": "NVIDIA GeForce RTX 4060 Ti",
            "platform": "gpu",
            "memory_bytes": 6438912000
          }
        ],
        "jax_x64_enabled": false
      },
      "static_memory_bytes": 879217408,
      "static_memory_gb": 0.819,
      "peak_memory_bytes": 2594863360,
      "peak_memory_gb": 2.417,
      "trace_path": "/fdtdx/tests/performance/results/20260702T065244Z_nvidia_geforce_rtx_4060_ti/traces/si_coupler_gds_cpl10",
      "memory_profile_path": "/fdtdx/tests/performance/results/20260702T065244Z_nvidia_geforce_rtx_4060_ti/memory_profiles/si_coupler_gds_cpl10.pb"
    }
  ]
}
```

## Summary by CodeRabbit

- **New Features**
  - Added a directional-coupler performance benchmark using FDTD simulation.
  - Supports configurable resolution, simulation steps, repetitions, tracing, and memory profiling.
  - Produces detailed performance, transmission, environment, and resource metrics in JSON format.
  - Added a coupler layout asset for benchmark simulations.

- **Chores**
  - Configured generated benchmark results to remain excluded from version control.


ymahlau:
Hi @teevee112, thanks for the PR. Without looking at the code in detail:

I like the idea of having a consistent and realistic performance measure that we can use to detect changes in execution speed. However, I do not think this should be set up as a test. Firstly, running a benchmark requires a GPU and some execution time, so we cannot do it on every PR. Second, between GPUs and even on a single GPU there is variability, so setting specific failure numbers in test cases will be a lot of effort to tune and maintain. Different jax versions also highly impact the speed. With a new jax version tracing might change and fdtdx becomes faster or slower. If it becomes faster the tests also become meaningless because they don't detect slowdowns anymore unless the thresholds are tuned again. 

For all these reasons, I think it would be better to create a single script parameterized by commandline arguments to run a speed benchmark. This could run the example simulation you used which i really like. 

If you want, you can create a top-level directory called 'performance' and add the gds file and the script in there. 

Some minor notes on the code while skimming through it:
- please do not change pyproject.toml or uv.lock
- keep the code size small and the script as simple as possible. It is easy to generate a lot of code with AI, but it quickly becomes difficult to maintain

teevee112:
@ymahlau reduced the scope of the PR by removing common utilities and extra functionality.

- Only one simulation included.
- The removed utilities gave us a common way to create result directories, capture environment metadata, export JSON, print summaries, wrap traced runs, collect memory profiles, and share scene setup across benchmarks.
  - We also lose automatic aggregation across variants. With pytest, it was straightforward to run several benchmark cases and get one combined result set. The standalone script is simpler, but it is less convenient if we add more cases later.
  - Trace debugging still works, but it is less turnkey. The script can still emit JAX traces with --trace; we are just not adding TensorBoard or a repo-managed trace-viewing path as part of this PR.
  - The important benchmark outputs remain: runtime, MCUPS, device/JAX info, JSON results, optional trace, and optional memory profile. Other plotting and debug utilities are excluded, to be added in the future.

Jinge-Wang:
> Hi @teevee112, thanks for the PR. Without looking at the code in detail:
> 
> I like the idea of having a consistent and realistic performance measure that we can use to detect changes in execution speed. However, I do not think this should be set up as a test. Firstly, running a benchmark requires a GPU and some execution time, so we cannot do it on every PR. Second, between GPUs and even on a single GPU there is variability, so setting specific failure numbers in test cases will be a lot of effort to tune and maintain. Different jax versions also highly impact the speed. With a new jax version tracing might change and fdtdx becomes faster or slower. If it becomes faster the tests also become meaningless because they don't detect slowdowns anymore unless the thresholds are tuned again.
> 
> For all these reasons, I think it would be better to create a single script parameterized by commandline arguments to run a speed benchmark. This could run the example simulation you used which i really like.
> 
> If you want, you can create a top-level directory called 'performance' and add the gds file and the script in there.
> 
> Some minor notes on the code while skimming through it:
> 
> * please do not change pyproject.toml or uv.lock
> * keep the code size small and the script as simple as possible. It is easy to generate a lot of code with AI, but it quickly becomes difficult to maintain

@ymahlau agree with the standalone-script direction. Two thoughts:

1. **Run before each release, store results outside the repo.** A benchmark run before every release (rather than per-PR) keeps the cost manageable. The JSON results with auto-logged env metadata (git SHA, JAX/jaxlib, driver, device are already captured by this PR) could live in a separate lightweight repo (e.g. `fdtdx-benchmarks`), so the main repo stays clean. Same for any production-scale GDS datasets a more comprehensive benchmark will eventually want, which will be too big and too niche for the main repo; they can be fetched on demand.

2. **Same-machine A/B as the primary regression tool.** Absolute MCUPS numbers are hostage to hardware/JAX/driver variability, and keeping them comparable over time needs pinned docker images and consistent cluster access, which is a lot of maintenance. But a ratio between two git refs measured back-to-back on the same machine is robust to all of it. A `--compare-ref <sha>` mode in the script (run base, run head, report the ratio) would give exactly the regression signal wanted, on whatever machine the person happens to have.

Optionally, a tiny `--perf-max-steps` invocation could run in CI (not to measure anything) just so the benchmark script itself doesn't bit-rot.

ymahlau:
@Jinge-Wang I like the idea of having a seperate repository for performance measurements. I added an issue for tracking this (see #410)

ymahlau:
@teevee112 thanks for adjusting this to a script! The PR looks good to me!

PR #399, opened 2026-07-09
Perf: optimize FDTD update hot paths

Zhenhao-Xie:
## Summary

Improves several FDTD update hot paths without changing the intended numerical behavior.

Changes include four parts:

- Curl update: replaces full-array `jnp.roll` curl stencils with direct neighbor slicing, and applies PML curl corrections component-wise before stacking the final curl array.
- PML update: keeps PML profiles in broadcast-shaped form instead of materializing full face arrays, and skips the zero-valued `kappa` correction when `kappa_start == kappa_end == 1`.
- Source update: adds a fast path for default always-on source updates, and updates axis-aligned point dipoles in scalar / diagonal media through only the active field component while keeping the full coupled path for rotated dipoles and full tensor media.
- Rectilinear grid: caches rectilinear-grid cell widths instead of recomputing `jnp.diff(edges)` on repeated access.

By reducing per-step dispatch overhead and unnecessary array materialization, especially for small and medium grids where GPU utilization was previously low, local GPU benchmarks on the same layered test case gave approximately:

- `0.8M` grid cells: `4.9x` faster
- `3.3M` grid cells: `2.7x` faster
- `7.4M` grid cells: `2.2x` faster

## Tests

Added / updated:

- Unit test coverage for the default always-on source-switch fast path.
- Unit test coverage for broadcast-shaped PML profiles.
- Unit test coverage ensuring full tensor point-dipole injection still keeps coupled field components.

Passed:

- `UV_CACHE_DIR=/tmp/uv-cache JAX_PLATFORMS=cpu uv run python -m pytest tests --cov --cov-branch --cov-config=pyproject.toml --cov-report=xml`
- Full local pytest coverage:
  - `2573 passed`
  - `1 skipped`
  - `1 xfailed`

## Summary by CodeRabbit

* **New Features**
  * Improved source handling for common “always on” configurations, reducing unnecessary overhead during field updates.
  * Added a faster path for axis-aligned dipole sources in compatible media, while preserving existing behavior for other cases.

* **Bug Fixes**
  * Improved curl and boundary calculations for more consistent simulation updates.
  * Fixed profile handling so boundary layers now return broadcast-friendly shapes and work correctly across grid dimensions.
  * Cached grid cell widths to avoid repeated recalculation during access.

### DIFF
--- a/src/fdtdx/core/grid.py
+++ b/src/fdtdx/core/grid.py
@@ -5,7 +5,7 @@ from matplotlib.path import Path
 
 from fdtdx import constants
 from fdtdx.core.axis import get_transverse_axes
-from fdtdx.core.jax.pytrees import TreeClass, autoinit, field, frozen_field, frozen_private_field
+from fdtdx.core.jax.pytrees import TreeClass, autoinit, field, frozen_field, frozen_private_field, private_field
 from fdtdx.core.misc import validate_symmetric_axis_cells
 
 
@@ -312,6 +312,7 @@ class RectilinearGrid(TreeClass):
     _min_spacings: tuple[float, float, float] = frozen_private_field()
     _is_uniform: bool = frozen_private_field()
     _uniform_spacing: float | None = frozen_private_field()
+    _cell_widths: tuple[jax.Array, jax.Array, jax.Array] = private_field(repr=False)
 
     def __post_init__(self):
         object.__setattr__(self, "x_edges", jnp.asarray(self.x_edges))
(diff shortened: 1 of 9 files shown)


Zhenhao-Xie:
@ymahlau I think CodeRabbit’s concern is valid. If someone updates `x_edges`, `y_edges`, or `z_edges` via `aset`, the new `_cell_widths` cache would become stale.

My understanding is that this follows the same assumption as the existing `_min_spacings`, `_is_uniform`, and `_uniform_spacing` caches, where grid edge arrays are treated as fixed after construction. I also did not find an internal path that mutates grid edges through `aset`.

I can add cache refresh logic if you prefer, but this feels like a broader cache-invalidation question that may be better handled in a separate PR, since several other caches in the codebase seem to rely on a similar assumption.

ymahlau:
Hi @Zhenhao-Xie, many thanks for this amazing PR! The caching mechanism is the same as in the rest of the repo, that's completely fine. 

I will accept the PR as is, I don't see any changes necessary. Just out of curiosity, did you find these levers for performance improvement yourself or did you ask some LLM to search for them?

PR #401, opened 2026-07-09
perf(mode): pin sliced mode-solver plane to single-device sharding

pic-asso:
## What

On multi-GPU runs, the field/material arrays (`inv_permittivities` / `inv_permeabilities`)
are x-sharded across the device mesh. `ModePlaneSource` and `ModeOverlapDetector` slice a
thin transverse plane out of these arrays and feed it to the Tidy3D mode-solver, which runs
as a `pure_callback` pinned to a single device.

Without an explicit sharding constraint on the sliced plane, XLA can conservatively
all-gather the **full x-sharded parent array** (~100M cells) to satisfy the callback's
single-device requirement, rather than gathering just the thin sliced plane (a few hundred
KB) it actually needs.

## How

Adds `pin_to_single_device` in `fdtdx/core/jax/sharding.py`: applies a `with_sharding_constraint`
that replicates the already-sliced plane across the full device mesh (not a plain
`SingleDeviceSharding`, which would raise "incompatible devices for jitted computation" against
the surrounding x-sharded computation). This tells XLA to gather only the small plane.

No-op on single-device runs (`len(jax.devices()) == 1`), so the single-GPU path is byte-identical
to before.

Applied at the two mode-solver call sites: `ModePlaneSource.place_on_grid` and
`ModeOverlapDetector.update`.

## Tests

`tests/unit/core/jax/test_sharding.py` (`TestPinToSingleDevice`, 4 new tests):
- single-device no-op returns the identical array object;
- multi-device: shape/dtype/values preserved (run in an isolated subprocess with 2 virtual
  CPU devices via `XLA_FLAGS=--xla_force_host_platform_device_count=2`, since the replicated
  path needs genuinely distinct physical devices — mocking `jax.devices()` with a duplicated
  device raises `ValueError: ... must be from distinct devices` on the eager
  `with_sharding_constraint` call);
- multi-device: result is a fully replicated `NamedSharding` (empty `PartitionSpec`);
- regression test for the real usage pattern — slicing a plane out of an x-sharded array
  inside `jax.jit` and pinning it must not raise the "incompatible devices" error.

All new tests verified to fail if the multi-device branch is stubbed out (sanity-checked
locally by temporarily forcing the no-op path unconditionally).

## Summary by CodeRabbit

* **New Features**
  * Added device pinning for sliced simulation data to ensure host callbacks receive only the intended slice in multi-device runs.

* **Bug Fixes**
  * Prevented unnecessary transfers/gathering of larger arrays when running mode-related computations.
  * Improved reliability of mode-solver callback outputs by aligning device placement to avoid sharding conversion errors.

* **Tests**
  * Added multi-device regression coverage (including callback and JIT scenarios) to validate correct replicated behavior and pinning semantics.

### DIFF
--- a/src/fdtdx/core/jax/sharding.py
+++ b/src/fdtdx/core/jax/sharding.py
@@ -41,6 +41,36 @@ def get_named_sharding_from_shape(
 counter = 0
 
 
+def pin_to_single_device(arr: jax.Array) -> jax.Array:
+    """Pin ``arr`` to a single-device replicated layout via ``with_sharding_constraint``.
+
+    On a multi-device run the field/material arrays are x-sharded (NamedSharding
+    ``P(None,'shard',None,None)``). When a thin transverse plane is sliced out of such an
+    array and fed to a host ``pure_callback`` (the Tidy3D mode solver, which runs at
+    ``{maximal device=0}``), XLA must materialise a single-device operand for the callback.
+    Without an explicit constraint XLA may conservatively all-gather the *parent* array to
+    satisfy the callback's single-device requirement. Pinning the already-sliced PLANE to a
+    single-device replicated sharding tells XLA to gather only the thin band (a few-hundred-KB
+    plane), never the full ~100M-cell domain.
+
+    No-op on a single-device run (``len(jax.devices()) == 1``): returns ``arr`` unchanged so the
+    single-GPU path stays byte-identical (no sharding constraint is inserted at all).
+    """
+    if len(jax.devices()) == 1:
+        return arr
+    # REPLICATED sharding over the FULL device mesh -- NOT SingleDeviceSharding, whose {device 0}
+    # device set is incompatible with the [0,1,...] set of the surrounding x-sharded computation
+    # (jit raises "Received incompatible devices for jitted computation"). A replicated
+    # PartitionSpec() keeps the thin sliced plane on the full mesh but UN-sharded, so XLA gathers
+    # only the small plane (a few-hundred KB) to materialise the callback operand, never the full
+    # x-sharded ~100M-cell domain, while staying device-set-compatible with the parent computation.
+    compute_devices = jax.devices()
+    devices = mesh_utils.create_device_mesh((len(compute_devices),), devices=compute_devices)
+    mesh = jax.sharding.Mesh(devices=devices, axis_names=(SHARD_STR,))
+    replicated = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec())
+    return jax.lax.with_sharding_constraint(arr, replicated)
+
+
 def get_dtype_bytes(dtype: jnp.dtype) -> int:
     return jnp.dtype(dtype).itemsize
 
--- a/src/fdtdx/core/jax/sharding.py
+++ b/src/fdtdx/core/jax/sharding.py
@@ -41,8 +41,26 @@ def get_named_sharding_from_shape(
 counter = 0
 
 
+def _replicated_mesh_sharding(compute_devices: list[jax.Device]) -> jax.sharding.NamedSharding:
+    """Build a REPLICATED NamedSharding over ``compute_devices`` -- NOT SingleDeviceSharding, whose
+    ``{device 0}`` device set is incompatible with the ``[0,1,...]`` set of the surrounding x-sharded
+    computation (jit raises "Received incompatible devices for jitted computation"). A replicated
+    PartitionSpec() keeps data on the full mesh but UN-sharded, so a subsequent placement (see
+    ``pin_to_single_device``) gathers only the small operand, never the full x-sharded domain, while
+    staying device-set-compatible with the parent computation.
+
+    Split out of ``pin_to_single_device`` so the (cheap, pure) mesh/sharding construction can be unit-
+    tested directly with a MOCKED device list (including a duplicated physical device, which the actual
+    data placement in ``pin_to_single_device`` cannot accept -- see ``jax.device_put``'s "distinct
+    devices" requirement, exercised only in the subprocess-based multi-device tests).
+    """
+    devices = mesh_utils.create_device_mesh((len(compute_devices),), devices=compute_devices)
+    mesh = jax.sharding.Mesh(devices=devices, axis_names=(SHARD_STR,))
+    return jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec())
+
+
 def pin_to_single_device(arr: jax.Array) -> jax.Array:
-    """Pin ``arr`` to a single-device replicated layout via ``with_sharding_constraint``.
+    """Pin ``arr`` to a single-device replicated layout via ``jax.device_put``.
 
     On a multi-device run the field/material arrays are x-sharded (NamedSharding
     ``P(None,'shard',None,None)``). When a thin transverse plane is sliced out of such an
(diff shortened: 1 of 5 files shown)


src/fdtdx/objects/sources/mode.py
```diff
@@ -112,10 +113,13 @@ def apply(
             raise NotImplementedError()
 
         # inv_permittivities shape: (3, Nx, Ny, Nz) - slice with component dimension
-        inv_permittivity_slice = inv_permittivities[:, *self.grid_slice]
+        # Pin the sliced PLANE to a single-device replicated layout (no-op on 1 GPU): the slice
+        # feeds the host mode-solver pure_callback (device 0), so this gathers only the thin band
+        # rather than letting XLA all-gather the full x-sharded parent array.
```
ymahlau: This comment is duplicated, a bit overdocumented

ymahlau:
Hi @pic-asso, thanks for the PR! I currently do not have easy access to a multi-gpu setup to verify this behavior. Could you provide some numbers / screenshot /logs or something similar to show that this PR does achieve what it claims?

Don't get me wrong, the reasoning and code all makes sense, its just that I currently have no way of verifying.

pic-asso:
Pushed a follow-up commit (`59c2ac5`) addressing the review feedback plus a real gap I found while benchmarking on real multi-GPU hardware.

## What's new in this commit

1. **The original fix only pinned the sliced *input* to the mode-solver's `pure_callback`, not the callback's own *outputs*.** `mode_E_raw`, `mode_H_raw`, and `eff_idx` come back from the host callback carrying a maximal/single-device `GSPMDSharding`, which is incompatible with the surrounding x-sharded computation. On 2 GPUs this still crashed inside `compute_mode`, just one step later than before. Now pinned via the same `pin_to_single_device`.
2. **`jax.lax.with_sharding_constraint` can't handle that specific sharding representation** — it raises `Cannot convert GSPMDSharding {maximal device=0} into SdyArray` internally via XLA's Shardy partitioner. Switched `pin_to_single_device`'s implementation to `jax.device_put`, which performs the placement directly and accepts both the already-`NamedSharding` sliced-input case and the maximal-sharded callback-output case.
3. Shortened the duplicated pinning comment in `sources/mode.py` / `detectors/mode.py` per the "overdocumented" nit.
4. Extracted `_replicated_mesh_sharding()` as a separately-testable pure function to close the patch-coverage gap — the mesh/sharding construction can now be unit-tested directly with a mocked device list, without needing the subprocess-based multi-device harness. Local coverage on `sharding.py` is now 96% (branch), with the remaining lines being the actual `device_put` placement call, which genuinely requires distinct physical devices to exercise.

## Benchmark (real 2×H100 hardware, non-PDK)

To validate this against real multi-GPU hardware without touching any confidential PDK geometry, I wrote a generic standalone benchmark: a straight strip waveguide (n_core=2.0 / n_clad=1.44, made-up numbers, no tie to any real platform), ~121M cells, with a `ModePlaneSource` feeding a downstream `ModeOverlapDetector` — same code path as `compute_mode`. The script monkey-patches `pin_to_single_device` to a no-op to simulate pre-#401 behavior on the same checkout, so all three phases below ran from the exact same commit.

Ran 3 phases in one SLURM allocation (1×H100 reference, 2×H100 without the callback-output pinning, 2×H100 with it):

| Phase | Overlap result | Peak GPU memory | Wall time |
|---|---|---|---|
| 1-GPU reference | `720.21387-111.2457j` | 18.3 GB (GPU0) | ~168s |
| 2-GPU, output pinning off | crashes at the callback-output sharding (the gap this commit fixes) | — | — |
| 2-GPU, output pinning on | `720.21387-111.2457j` | 17.3 GB (GPU0) + 13.2 GB (GPU1) | ~171s |

Bit-exact correctness (identical overlap to full float32 precision) between 1-GPU and the fully-patched 2-GPU run — that, plus no longer crashing, is the actual point of this fix.

I'd flag the memory/timing columns above as **inconclusive on their own**: GPU0 only drops 18.3→17.3GB (~5%), the 2-GPU *total* (30.5GB) is actually higher than the 1-GPU total (18.3GB), and the split is uneven (17.3 vs 13.2) — at this size the fixed overhead (mesh setup, program binary, small non-sharded state) is a large enough fraction of the total that it can obscure a real halving one way or the other. Wall-clock is flat-to-slightly-worse too (~168s vs ~171s).

To get a cleaner read, I re-ran the same scene ~1.6× larger (NX 2500→4000, ~194M cells instead of ~121M):

| Phase | Peak GPU memory | Wall time |
|---|---|---|
| 1-GPU reference | 19.6 GB (GPU0) | ~460s |
| 2-GPU, patched | 17.5 GB (GPU0) + 17.5 GB (GPU1) | ~467s |

Here the 2-GPU split is genuinely even (17.51 vs 17.51 GB, not 17.3 vs 13.2) and both devices sit below the 1-GPU figure — a much more convincing sharded-not-replicated signature, even though the wall-clock still doesn't improve (same story: fixed per-step sync + the serial callback dominate). I don't want to overclaim a big memory win from either benchmark, but the larger one is a materially cleaner demonstration than the original table, so pairing them here for completeness.

**Edit:** removed a closing claim here about `PhasorDetector` needing separate handling — a follow-up reproduction showed that was wrong (a scene with a `PhasorDetector` runs cleanly under this fix; the crash I'd seen earlier was this same callback-output issue, not `PhasorDetector`). No action needed there.

**Final wrap-up:** while chasing the above I also found that gradient-mode multi-GPU (both `method="reversible"` and `"checkpointed"`) doesn't shard its backward-pass state at all — replicates the full buffer on every device instead. That's unrelated to this PR's scope (mode-solver `pure_callback` sharding) and doesn't touch `PhasorDetector` either, so I filed it separately as #417 rather than folding it in here. From my side this PR is done — the fix, tests, and benchmarks above are the full extent of what I have.

src/fdtdx/core/jax/sharding.py
```diff
@@ -41,6 +41,60 @@ def get_named_sharding_from_shape(
 counter = 0
 
 
+def _replicated_mesh_sharding(compute_devices: list[jax.Device]) -> jax.sharding.NamedSharding:
+    """Build a REPLICATED NamedSharding over ``compute_devices`` -- NOT SingleDeviceSharding, whose
```
ymahlau, line 45: Please remove all of the AI chain-of-thoughts comments. We want to have actual docstrings, not a development story

src/fdtdx/objects/sources/mode.py
```diff
@@ -112,10 +113,11 @@ def apply(
             raise NotImplementedError()
 
         # inv_permittivities shape: (3, Nx, Ny, Nz) - slice with component dimension
-        inv_permittivity_slice = inv_permittivities[:, *self.grid_slice]
+        # Pin the sliced plane to a single device (see pin_to_single_device docstring for why).
```
ymahlau, line 116: references to other comments are generally not good style

tests/unit/core/jax/test_sharding.py
```diff
@@ -198,3 +204,184 @@ def test_sharding_axis_fallback_when_dim_is_one(self):
     def test_has_named_sharding(self):
         result = create_named_sharded_matrix(shape=(4, 6), value=1.0, sharding_axis=0, dtype=jnp.float32, backend="cpu")
         assert isinstance(result.sharding, jax.sharding.NamedSharding)
+
+
+# ---- _replicated_mesh_sharding ----
```
ymahlau, line 209: remove excessive comments and unnecessary spacing

tests/unit/core/jax/test_sharding.py
```diff
@@ -198,3 +204,184 @@ def test_sharding_axis_fallback_when_dim_is_one(self):
...
+
+
+class TestReplicatedMeshSharding:
+    """Tests for the _replicated_mesh_sharding helper, split out of pin_to_single_device so this
```
ymahlau, line 213: This comment is also convoluted

tests/unit/core/jax/test_sharding.py
```diff
@@ -198,3 +204,184 @@ def test_sharding_axis_fallback_when_dim_is_one(self):
...
+        assert len(jax.devices()) == 1
+        arr = jnp.arange(12.0).reshape(3, 4)
+        result = pin_to_single_device(arr)
+        assert result is arr
+
+    def test_multi_device_preserves_shape_dtype_and_values(self):
+        _run_in_subprocess_with_devices(
+            2,
+            """
+            import jax, jax.numpy as jnp
```
ymahlau, line 290: Is there any way to run these unit tests without having python code as strings? This kind of code is hard to maintain, maybe a better variant would be defining a helper function directly as python code and executing that

ymahlau:
Hi @pic-asso, the `pre-commit` checks are currently failing an the PR needs some re-work to remove the AI bloat. From a functional point of view this looks good to me

PR #402, opened 2026-07-10
feat: per-axis pole parameters for diagonally anisotropic dispersion

MarkCunningham0410:
This adds per-axis pole parameters to the dispersion models, enabling diagonally anisotropic dispersion, and with it natural hyperbolic materials (hBN and friends), where Re(ε) is negative along some crystal axes and positive along others inside a Reststrahlen band. Since a static negative permittivity is unconditionally unstable in explicit FDTD, such materials can only be modelled through dispersion, which so far applied one pole set to all three field components.

Every pole parameter now accepts either a scalar (isotropic, behaviour unchanged) or an (x, y, z) tuple, following the parameter-widening pattern from #75:

```
# hBN: different phonon resonances in-plane vs out-of-plane
hbn = fdtdx.Material(
    permittivity=(4.87, 4.87, 2.95),  # eps_inf per axis
    dispersion=fdtdx.DispersionModel(poles=(
        fdtdx.LorentzPole(
            resonance_frequency=(w_to_perp, w_to_perp, w_to_par),
            damping=(g_perp, g_perp, g_par),
            delta_epsilon=(de_perp, de_perp, de_par),
        ),
    )),
```
A pole acting on a single axis is expressed by zeroing its strength on the others, e.g. DrudePole(plasma_frequency=(wp, 0.0, 0.0), damping=g) for a medium that is metallic only along x.

## Implementation notes:

- The dispersive coefficient arrays get a conditional component axis: `(num_poles, 1, ...)` as before when all dispersion is isotropic (no memory or behaviour change), `(num_poles, 3, ...)` when any material has per-axis poles — mirroring the existing 1/3/9 material array sizing.
- The E-field update and its time-reversed counterpart are already elementwise over that axis, so the update equations are untouched (comment changes only). Reversible autodiff works unchanged.
- New `compute_pole_coefficients_per_axis` returns `(n_poles, 3)` coefficients; the existing `compute_pole_coefficients` and the scalar pole accessors are bit-identical for isotropic poles and raise a descriptive error for per-axis ones. `DispersionModel.susceptibility_axes` / `permittivity_axes` evaluate per-axis spectra.
- `Material` now warns when a static (non-dispersive) diagonal permittivity entry is ≤ 0, since that diverges silently at runtime.

## Two fixes found while testing:

- Dispersion combined with a fully anisotropic (off-diagonal) conductivity tensor silently skipped the ADE update - the full-tensor update branch has no polarisation step and the existing guard only checked the permittivity. This now raises `NotImplementedError` like the permittivity case.
- `has_dispersive_edot` and `compute_eps_spectrum_from_coefficients` assumed the size-1 component axis. The broadband TFSF impedance filter now also logs a warning in per-axis-dispersive media (it uses a component-averaged ε(ω); the carrier-frequency correction stays exact per component).

## Validation:

- A per-axis Lorentz slab matches the equivalent isotropic run per polarization (the arithmetic is identical, the coefficients just stop broadcasting).
- x/y-polarised transmission through a per-axis Lorentz half-space matches Fresnel computed from the per-axis susceptibility, at the same tolerance as the existing isotropic dispersion tests.
- Hyperbolic sanity check: a per-axis Drude half-space with Re(ε_x) < 0 < ε_y reflects x-polarization like a metal and transmits y-polarization per Fresnel.
- Reversible and checkpointed gradients agree (float64, rel < 1e-5) with per-axis coefficients; the time-reversal round-trip reconstructs E/H and the polarization state exactly.
- Full unit + integration suite passes; isotropic setups produce identical arrays.

Example: `examples/hyperbolic_dispersive_slab.py` — per-axis Drude, ε(ω) plot per axis, polarization-selective transmission through the hyperbolic band.

Out of scope: dispersion with off-diagonal (rotated) tensors, the existing `NotImplementedError` stays in place (and now also covers off-diagonal σ). Optical axes must align with the grid.

Related: #4, #75, #204, #294.

## Summary by CodeRabbit

* **New Features**
  * Added support for diagonally anisotropic dispersive materials with per-axis Lorentz, Drude, and CCPR parameters.
  * Added public per-axis susceptibility, permittivity, and pole-coefficient calculations.
  * Added an example demonstrating hyperbolic dispersive media and polarization-selective transmission.

* **Bug Fixes**
  * Improved validation and warnings for unstable permittivity and unsupported anisotropic conductivity combinations.
  * Added support for per-axis dispersion across material initialization, interpolation, and simulations.

* **Documentation**
  * Updated API documentation, usage guidance, and shape descriptions for per-axis dispersion.

### DIFF
--- a/src/fdtdx/dispersion.py
+++ b/src/fdtdx/dispersion.py
@@ -48,6 +48,19 @@ Stability requires :math:`\\gamma \\Delta t < 2`; in the physical regime
 of interest :math:`\\gamma \\Delta t \\ll 1`, so :math:`c_2 \\approx -1`,
 which makes the backward (reverse-time) recurrence numerically stable
 after algebraic inversion.
+
+Anisotropic (per-axis) dispersion
+---------------------------------
+Every pole parameter accepts either a scalar (isotropic, applied to all
+three axes) or a 3-tuple ``(x, y, z)`` giving a different value per grid
+axis. This yields a diagonally anisotropic susceptibility tensor
+:math:`\\chi(\\omega) = \\mathrm{diag}(\\chi_x, \\chi_y, \\chi_z)` — enough to
+model uniaxial/biaxial crystals and hyperbolic media (e.g. hBN) whose
+optical axes align with the grid. A pole that only acts on one axis is
+expressed by zeroing its strength on the others, e.g.
+``LorentzPole(resonance_frequency=w0, damping=g, delta_epsilon=(2.25, 0.0, 0.0))``:
+with zero coupling the polarization on that axis stays identically zero.
+Off-diagonal (rotated) dispersive tensors are not supported.
 """
 
 from __future__ import annotations
@@ -62,6 +75,24 @@ from fdtdx.constants import eps0
 from fdtdx.core.jax.pytrees import TreeClass, autoinit, frozen_field
 
 
+def _broadcast_axis_param(value: float | complex | tuple) -> tuple:
+    """Normalize a pole parameter to a per-axis 3-tuple ``(x, y, z)``.
+
+    Scalars are broadcast to all three axes; 3-tuples pass through unchanged.
+    """
+    if isinstance(value, tuple):
+        if len(value) != 3:
+            raise ValueError(
+                f"Per-axis pole parameters must be a scalar or a 3-tuple (x, y, z), got a tuple of length {len(value)}."
+            )
+        return value
+    return (value, value, value)
+
+
+def _is_uniform(axes: tuple) -> bool:
+    return bool(axes[0] == axes[1] == axes[2])
+
+
 @autoinit
 class Pole(TreeClass, ABC):
     """Abstract base class for a single 2nd-order ADE pole.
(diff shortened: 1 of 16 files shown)


ymahlau:
Awesome @MarkCunningham0410 ! I will probably not have time to look at this today, will take a closer look next week

src/fdtdx/dispersion.py
```diff
@@ -519,15 +735,18 @@ def compute_eps_spectrum_from_coefficients(
...
     numer = a_dt2[None, ...] - 1j * omega_dt * b_dt[None, ...]
     denom = omega0_sq_dt2[None, ...] - omega_dt**2 - 1j * gamma_dt[None, ...] * omega_dt
     safe_denom = np.where(pole_mask[None, ...], denom, 1.0 + 0.0j)
     chi_per_pole = np.where(pole_mask[None, ...], numer / safe_denom, 0.0 + 0.0j)
-    chi_per_cell = chi_per_pole.sum(axis=1)  # sum over pole axis → (M, 1, *spatial)
-    chi_per_cell = chi_per_cell[:, 0]  # drop component broadcast axis → (M, *spatial)
+    chi_per_cell = chi_per_pole.sum(axis=1)  # sum over pole axis → (M, C, *spatial)
+    # Average the material-component axis (identity for C = 1), mirroring the
+    # eps_inf reduction above — this scalar spectrum feeds an impedance filter
+    # that has no notion of polarization.
```
ymahlau, line 748: Could you add a TODO here that the average is a choice for initial implementation and correct would be some polarization-weighting (if i understand correctly)?

ymahlau:
Thanks @MarkCunningham0410 for this strong PR! I looked through the code and basically have no comments for improvements. 

I added a follow-up issue (see #412) to track also enabling polarization-weighted averages to make sources in anisotropic dispersive media work correctly.

MarkCunningham0410:
Ah amazing! Thanks @ymahlau!

I spent a bit of the weekend working on getting twisted anisotropic materials working too (and tilts etc for an arbitrary 3x3 dielectric tensor) so hopefully a PR coming on that soon!

PR #404, opened 2026-07-10
Circumvent tidy3d's backward mode computation

l-berg:
tidy3d's mode solver recently introduced a normalization step which shifts the phase by +-i when `direction="-"`. This means that the transversal fields are no longer real. This breaks the `direction="-"` of our mode sources and detectors. I raised this in their repo under flexcompute/tidy3d#3379.

This PR is intended as a workaround on our side. We handle the direction change ourselves and always pass `direction="+"` to tidy3d's solver.

# Example

Reversing the direction of the source in the example notebook from the docs results in these fields.

## direction = "+"
Screenshot 2026-07-10 190157

## direction = "-"
Screenshot 2026-07-10 194853

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backward waveguide mode calculations using consistent reciprocity-based field transformations.
  * Corrected backward-mode phase conventions and field signs for supported material configurations.
  * Added validation and clear errors for unsupported angled or non-reciprocal material cases.

* **Tests**
  * Added coverage for isotropic, anisotropic, and tensorial waveguide materials.
  * Verified effective indices, field transformations, phase behavior, and backward power flow.

### DIFF
--- a/src/fdtdx/core/physics/modes.py
+++ b/src/fdtdx/core/physics/modes.py
@@ -377,6 +377,15 @@ def compute_mode(
     return mode_E_norm, mode_H_norm, eff_idx
 
 
+def _is_reciprocal_tensor(components: Sequence[ArrayLike], tol: float = 1e-6) -> bool:
+    """Check whether a material tensor given as 9 row-major components is symmetric (reciprocal)."""
+    return bool(
+        np.max(np.abs(np.asarray(components[1]) - np.asarray(components[3]))) <= tol
+        and np.max(np.abs(np.asarray(components[2]) - np.asarray(components[6]))) <= tol
+        and np.max(np.abs(np.asarray(components[5]) - np.asarray(components[7]))) <= tol
+    )
+
+
 def tidy3d_mode_computation_wrapper(
     frequency: float,
     permittivity_cross_section: ArrayLike,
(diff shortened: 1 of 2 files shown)


Jinge-Wang:
@ymahlau as I am skimming through the relevant PRs and issues, I wonder if there is a general interest in a **native** mode solver here as a default engine? We can keep tidy3D as an optional backend for cross-checks and the hard cases (e.g. fully tensorial media, bends ...). A full-vectorial finite-difference solver on the 2D Yee cross-section is well documented: Zhu & Brown, Opt. Express 10, 853 (2002), and Fallahkhair, Li & Murphy, J. Lightwave Technol. 26, 1423 (2008), and solving it as a sparse shift-invert eigenproblem near the target effective index is straightforward with scipy. That would make incidents like the upstream normalization change here a non-event, and would drop a heavy dependency for the common straight-waveguide case. 

Happy to write up a concrete proposal if there's interest.

ymahlau:
Hi @Jinge-Wang, thanks for your interest and offer to help! We basically already finished the development of an internal mode solver (with fully tensorial, bends, automatic differentiation, ...), but still need thorough testing and are planning to write a publication on the autodiff parts. It will be integrated in the next months together with the efforts of #349.

Jinge-Wang:
@ymahlau That's sounds exciting. Really appreciate the work you've been putting into this and looking forward to the integration and the publication.

momchil-flex:
As commented [here](https://github.com/flexcompute/tidy3d/issues/3379), the `EigSolver` was never meant to produce a deterministic gauge, but I agree that there's definitely a confusion about whether it's supposed to be publicly used or not.

The fix here seems good to restore the old contract that fdtdx had with tidy3d; but more generally you guys might want to set a phase explicitly (apart from normalization), or use our `ModeSolver` rather than the internal `EigSolver`.

ymahlau:
@momchil-flex thanks for the clarification, we will use this fix until the development / testing of an internal mode solver is complete. FYI I think the differentiable mode solver we are developing could also be interesting for tidy3d

momchil-flex:
It can certainly be interesting, is it also jax or how do you handle the eigensolve backprop?

Oh and the most important question: does it support autodiff through the mode fields or only the frequencies?

For the longest time I was receiving [updates on this issue](https://github.com/jax-ml/jax/issues/2748) which was opened in 2020 and I participated in a bit. So I know it got closed recently and presumably the eigenvector support is stable-ish now? :D I can imagine things are still tricky if modes are (close to) degenerate.

ymahlau:
The main solver is a custom adjoint on the mode solve supporting both mode fields and frequencies. We are still in the process of validating everything and degenerate modes are certainly difficult to handle. This part is written in numpy/scipy with a thin jax/pytorch/autograd wrapper. So we are not even using the jax issue pointed out above :)

We also implemented the matrix-free solver of Grey et al. [see [here](https://opg.optica.org/oe/fulltext.cfm?uri=oe-32-17-30541)], which runs in jax simply because that is faster. But since we are planning to move fdtdx towards pytorch anyways that might also change.

PR #406, opened 2026-07-12
Perf: region-restricted detector recording + opt-in phasor DFT subsampling

Jinge-Wang:
## Summary

Detector recording currently interpolates E/H over the **whole domain on every time step** unconditionally, even when detectors (1) are off at that step, (2) don't use exact interpolation, or (3) only cover a small monitor plane. Each detector then re-slices the full-domain arrays inside its own `update()`.

This PR restricts all detector recording to each detector's own region and gates it on the detector's activity switch:

- `update_detector_states` now passes fields and materials **already restricted to `grid_slice`** into `Detector.update()`.
- A strictly interior detector with `exact_interpolation` is co-located over just its region plus a one-cell halo, **inside** the `lax.cond` gate (skipped entirely on off-steps). This is byte-identical to slicing the full-domain interpolation (see parity test). The H time-centering `(H_prev + H) / 2` is likewise computed only on the detector block.
- The full-domain interpolation is built only as a shared fallback for exact detectors whose stencil touches a domain edge (where boundary padding matters), preserving existing behavior there.
- On non-uniform `RectilinearGrid`s, a new optional `region_slice` argument to `interpolate_fields` keeps the physical-distance weights aligned with the sub-block.
- New opt-in `PhasorDetector.dft_subsample` (`int` or `"auto"`): records only every stride-th active step with the kept samples rescaled, exploiting the fact that FDTD time steps oversample optical frequencies by ~10–20×. `"auto"` keeps 12 samples per period of the highest recorded frequency; explicit strides near Nyquist trigger a warning. Default (1) is byte-identical to current behavior.

## Performance

Forward run, PML, Gaussian plane source, w/ three always-on monitors (24×24 Poynting plane, 24×24 phasor plane, 16^3 reduced energy volume), 629 steps, float32, CPU (Apple Silicon):

| Domain | Detector overhead (main) | Detector overhead (this PR) | Reduction |
|---|---|---|---|
| 72^3 | 0.356 ms/step | 0.165 ms/step | **2.2×** |
| 96^3 | 0.735 ms/step | 0.292 ms/step | **2.5×** |

End-to-end runtime with this monitor set drops ~12% at both sizes; the gap widens with domain size because the old cost scales with the domain volume while the new cost scales with monitor size. `dft_subsample="auto"` reduces phasor recording cost by a further ~2–10× depending on resolution (validated to 0.16% agreement in a full simulation, see tests).

## Breaking change

`Detector.update()` implementations now receive region-restricted inputs and **must not slice `self.grid_slice` themselves**. All built-in detectors are updated; downstream custom detectors (ones outside fdtdx repo) need the same one-line deletion. Two safeguards report the failure into an actionable error:

- exceptions raised inside `update()` are re-raised naming the detector and the double-slicing cause;
- a trace-time check rejects updates whose returned state layout drifted from `init_state()`.

Also fixes a latent inconsistency found while consolidating: box-mode `FieldProjectionDetector` duplicated the phasor scaling logic; both paths now share `PhasorDetector._static_scale()` (fixes missing pulse-mode stride weight).

The stale `update_detector_states` docstring ("interpolation is disabled by default") is corrected to match the actual default (`exact_interpolation=True` since 7a05ac1) <- this default description was left unchanged after that update.

## Testing

- `tests/integration/fdtd/test_detector_region_parity.py` — byte-level parity of region recording vs. the historical full-domain path for every detector kind, on uniform and non-uniform grids.
- `tests/unit/core/physics/test_curl.py` — `region_slice` sub-block interpolation equals the sliced full-domain result on both grid types.
- `tests/unit/objects/detectors/test_phasor_subsample.py` — stride resolution, scaling relations, single-tone amplitude recovery, Nyquist warning.
- `tests/simulation/fdtd/test_detector_region.py` — gradient through `reversible_fdtd` is finite and nonzero via a region-recorded detector; `dft_subsample="auto"` matches every-step recording in a full simulation (0.16% max deviation, asserted < 5%).
- `tests/unit/fdtd/test_update.py` — rewritten `TestUpdateDetectorStates` covering the interior/edge/non-exact paths, material slicing, and both safeguards.

Full suite: 2492 passed (`-m "unit or integration or docs"`), pre-commit (ruff/ty/zizmor) and `sphinx-build -W` docs build clean.

## Some future considerations for performance 

- Defer co-location interpolation for the phasor family by accumulating the DFT of raw staggered fields and interpolating the accumulated phasor once post-run (linear ops commute; same design as Meep's `add_dft_fields`).
- Formalize a postprocess stage for detector data (the natural home for the above).

## Summary by CodeRabbit

* **New Features**
  * Added configurable phasor DFT subsampling with automatic stride selection and Nyquist/aliasing warnings.
  * Improved detector recording to handle region-restricted updates more robustly, including edge-aware behavior.
  * Extended field interpolation to support non-uniform rectilinear grid scaling and region-slice interpolation.

* **Bug Fixes**
  * Improved detector state validation with clearer mismatch errors and context.
  * Corrected detector input handling/scaling for subsampled pulse modes and detector-specific data extraction.

* **Tests**
  * Added/expanded integration and unit tests for region parity, non-uniform interpolation correctness, subsampling behavior, and reversible autodiff gradients.

### DIFF
--- a/src/fdtdx/core/physics/curl.py
+++ b/src/fdtdx/core/physics/curl.py
@@ -4,6 +4,7 @@ import jax.numpy as jnp
 from fdtdx.config import SimulationConfig
 from fdtdx.constants import c as c0
 from fdtdx.fdtd.container import ObjectContainer, PmlAuxField
+from fdtdx.typing import SliceTuple3D
 
 
 def _metric_scale(
@@ -43,12 +44,24 @@ def _backward_edge_average(
     previous: jax.Array,
     config: SimulationConfig | None,
     axis: int,
+    region_slice: SliceTuple3D | None = None,
 ) -> jax.Array:
     """Interpolate center-staggered samples back to an edge on a rectilinear grid.
 
     Uniform grids use the historical arithmetic mean.  On stretched grids the
     target edge is not halfway between neighboring cell centers, so the average
     is weighted by the half-widths of the cells on each side of the edge.
+
+    Args:
+        current: Center-staggered samples of the cells on the target edges.
+        previous: Center-staggered samples of the neighboring cells behind the edges.
+        config: Optional simulation configuration carrying the grid.
+        axis: Grid axis along which the interpolation is performed.
+        region_slice: Grid slice ``((x0, x1), (y0, y1), (z0, z1))`` covered by the inputs
+            when they are a sub-block of the domain. The cell widths are sliced to match.
+
+    Returns:
+        Edge-interpolated samples of the same shape as ``current``.
     """
     if config is None or not config.has_nonuniform_grid:
         return 0.5 * (current + previous)
@@ -57,6 +70,10 @@ def _backward_edge_average(
     assert grid is not None
     widths = grid.cell_widths(axis)
     previous_widths = jnp.concatenate([widths[:1], widths[:-1]])
+    if region_slice is not None:
+        start, stop = region_slice[axis]
+        widths = widths[start:stop]
+        previous_widths = previous_widths[start:stop]
     current_half_width = 0.5 * widths
     previous_half_width = 0.5 * previous_widths
     broadcast_shape = [1, 1, 1]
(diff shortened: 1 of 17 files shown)


ymahlau:
Thanks @Jinge-Wang for this great PR. I have no major comments, this can go in as is :)

PR #407, opened 2026-07-12
fix: apply curl metric factor to TFSF injection on non-uniform grids

### LINKED ISSUE #392: Potential Bug: QuasiUniformGrid TFSF Source Normalization
## Summary

`TFSFPlaneSource` / `UniformPlaneSource` may over-inject source amplitude on anisotropic `QuasiUniformGrid`.

The bulk curl update is metric-aware, but the TFSF source correction still appears to use only scalar `config.courant_number`. For a `+z` source with `dx=dy=6.25 nm`, `dz=12.5 nm`, this predicts about `3x` too much injected power, close to the observed `~2.8x` excess in `Sz`.

## Environment

- FDTDx: editable install from `main`
- Commit tested: `9494b0e`
- Grid: `QuasiUniformGrid(dx=6.25e-9, dy=6.25e-9, dz=12.5e-9)`
- Source: normal-incidence `UniformPlaneSource`, `direction="+"`, Ex polarization

## Evidence

Bulk curl uses per-axis metric scaling:

```python
reference_spacing = c0 * config.time_step_duration / config.courant_number
scale = reference_spacing / widths
```

But `TFSFPlaneSource.update_E/update_H` uses:

```python
c = self._config.courant_number
```

without an analogous propagation-axis metric factor.

For this grid:

```text
reference_spacing = 7.2169 nm
z scale = reference_spacing / dz = 0.57735
predicted amplitude excess = 1 / 0.57735 = 1.732
predicted power excess = 1.732^2 = 3.0
observed mean Sz excess = ~2.83
```

## Question

Is this a bug in `TFSFPlaneSource` for `QuasiUniformGrid`?

Should `TFSFPlaneSource.update_E/update_H` multiply source corrections by the propagation-axis metric factor, e.g.

```python
source_metric_scale = (c0 * dt / courant_number) / d_propagation
```

matching the metric-aware curl update?

Jinge-Wang:
## Summary

Fixes #392. On non-uniform grids (`QuasiUniformGrid` / `RectilinearGrid`), the bulk curl updates scale each spatial derivative by the local metric factor `reference_spacing / cell_width` (`_metric_scale` in `core/physics/curl.py`), but the TFSF source correction in `TFSFPlaneSource.update_E` / `update_H` still injected with the bare scalar `courant_number`. Since the TFSF correction is a curl term at the source plane, it must carry the same local factor along the propagation axis. Without it, a source on a grid with dz != dx over-injects power by roughly `(dz/dx)^2`.

## Fix

Adds `TFSFPlaneSource._metric_scale_at_plane(stencil)`, mirroring the curl's `_metric_scale` for the single cell of the source plane, and multiplies the injected corrections by it:

- `update_E` uses the **backward** stencil (matching `curl_H`, which the E update consumes),
- `update_H` uses the **forward** stencil (matching `curl_E`).

On uniform grids the factor is exactly 1.0 and the behavior is unchanged. All TFSF-derived sources (`UniformPlaneSource`, `GaussianPlaneSource`, `LinearlyPolarizedPlaneSource`, `ModePlaneSource`) inherit the fix. The forward/backward (inverse) injection paths share the same code, so gradient runs stay consistent.

## Validation

Reproduction from the issue (vacuum, +z `UniformPlaneSource`, same physical domain and detector plane, PML z / periodic xy, steady-state flux through a fixed physical area):

| Grid | Flux ratio vs uniform (before) | (after) |
|---|---|---|
| `QuasiUniformGrid(dx=dy=25nm, dz=50nm)` | **2.98x** | **0.99** |

The above matches the ~3x predicted and ~2.8x observed in #392 (before fix).

New regression test `tests/simulation/physics/sources/test_tfsf_grid_normalization.py` asserts injected plane-wave power is grid-metric independent (ratio within 5%). Full unit+integration suite (2473 tests), pre-commit, and ty pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved total-field/scattered-field plane-wave source normalization on non-uniform grids by applying local metric scaling to the Courant-based coefficients.

* **Tests**
  * Added regression tests verifying injected wave power is consistent across uniform, quasi-uniform, and graded propagation-axis grids.
  * Added unit coverage for the new plane metric-scaling behavior, including edge handling and invalid stencil errors.

### DIFF
--- a/src/fdtdx/objects/sources/tfsf.py
+++ b/src/fdtdx/objects/sources/tfsf.py
@@ -5,6 +5,7 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 
+from fdtdx.constants import c as c0
 from fdtdx.core.jax.pytrees import autoinit, frozen_field, private_field
 from fdtdx.dispersion import (
     compute_eps_spectrum_from_coefficients,
@@ -307,6 +308,33 @@ class TFSFPlaneSource(DirectionalPlaneSourceBase, ABC):
         # the real impedance of the local medium.
         raise NotImplementedError()
 
+    def _metric_scale_at_plane(self, stencil: str) -> jax.Array | float:
+        """Computes the local derivative scale along the propagation axis at the source plane.
+
+        On non-uniform grids, spatial derivatives are scaled by the local cell width.
+        Returns 1.0 on uniform grids.
+
+        Args:
+            stencil (str): Either "backward" or "forward" difference stencil.
+
+        Returns:
+            jax.Array | float: Scalar scale factor for the injected field corrections.
+        """
+        config = self._config
+        if not config.has_nonuniform_grid:
+            return 1.0
+        grid = config.resolved_grid
+        assert grid is not None
+        widths = grid.cell_widths(self.propagation_axis)
+        start = self.grid_slice_tuple[self.propagation_axis][0]
+        width = widths[start]
+        if stencil == "backward":
+            width = 0.5 * (width + widths[max(start - 1, 0)])
+        elif stencil != "forward":
+            raise ValueError(f"Unknown derivative stencil: {stencil}")
+        reference_spacing = c0 * config.time_step_duration / config.courant_number
+        return reference_spacing / width
+
     def update_E(
         self,
         E: jax.Array,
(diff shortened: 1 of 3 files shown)


ymahlau:
Hi @Jinge-Wang, thanks for so quickly making a PR to fix the issue! No review comments from my side

PR #409, opened 2026-07-13
Fix stale package version in JsonSetup exports

seansongss:
## Summary

`JsonSetup.dumps()` currently hard-codes the exported FDTDX version as `0.6.0`, while the package version is `0.6.2`. This can cause newly generated JSON files to contain incorrect version metadata.

This PR:

- Retrieves the installed FDTDX version dynamically instead of hard-coding it.
- Uses the current package version for the `__fdtdx_version__` field.
- Adds a test to verify that exported JSON contains the correct version.

This prevents the JSON metadata from becoming outdated when new package versions are released.

## Tests

Added and ran the relevant JSON conversion tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Exported JSON data now reports the installed package version accurately instead of relying on a fixed value.

* **Tests**
  * Added validation to ensure serialized JSON includes the correct package version.

### DIFF
--- a/src/fdtdx/conversion/json.py
+++ b/src/fdtdx/conversion/json.py
@@ -2,6 +2,7 @@ import dataclasses
 import importlib
 import json
 from dataclasses import dataclass
+from importlib.metadata import version
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence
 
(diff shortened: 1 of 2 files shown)


ymahlau:
@seansongss thanks for catching this, PR looks good to me!

PR #428, opened 2026-07-29
feat: temporal apodization windows for PhasorDetector

bruxillensis:
Adds an optional `apodization` window to `PhasorDetector`, applied to each sample of its running DFT. The default (`None`) is the existing hard rectangular gate, so nothing changes unless it is set.

A rectangular gate leaks spectrally: a signal still ringing when the detector switches off hits a discontinuity, and the sidelobes land on neighbouring frequencies. Measuring a feature 1000x below a nearby strong tone, the gate overstates it by ~60x; a Tukey window brings it to the right order.

Windows live in `core/window.py` alongside the envelope helpers, which are extracted from the inline copies in `objects/sources/profile.py` so the source profiles and the windows share one definition of each shape. `WindowProfile` takes `get_window(time)`, with `apodization` typed to it so a source profile is rejected at construction.

Continuous-mode scaling becomes the window's coherent gain `2 / sum(w)`, which reduces to the previous `2 / N` for the rectangular case. The sum is taken over the recorded steps, so this composes with `dft_subsample`.

## Summary by CodeRabbit

* **New Features**
  * Added shared temporal window/envelope utilities and new `WindowProfile` implementations: `GaussianWindowProfile` and `TukeyWindowProfile`.
  * Enabled optional temporal apodization for `PhasorDetector` via `WindowProfile`, including coherent-gain correction for continuous recordings.
* **Documentation**
  * Extended API docs to include `WindowProfile`, `GaussianWindowProfile`, and `TukeyWindowProfile`.
* **Refactor**
  * Updated source temporal shaping to use shared Gaussian and linear ramp helpers for smoother transitions.
* **Tests**
  * Added unit tests for envelope/window shapes, parameter validation, apodization edge cases (including rejection for invalid/underflowed windows), and detector subsampling behavior.
  * Added physics regression coverage to confirm reduced spectral leakage off DFT bins with Tukey windows.

### DIFF
--- a/docs/source/07_api.rst
+++ b/docs/source/07_api.rst
@@ -67,6 +67,7 @@ API
     fdtdx.GaussianPlaneSource
     fdtdx.GaussianPulseProfile
     fdtdx.GaussianSmoothing2D
+    fdtdx.GaussianWindow
     fdtdx.GradientConfig
     fdtdx.GridCoordinateConstraint
     fdtdx.HorizontalSymmetry2D
@@ -130,8 +131,10 @@ API
     fdtdx.SubpixelSmoothedProjection
     fdtdx.TanhProjection
     fdtdx.TemporalProfile
+    fdtdx.TemporalWindow
     fdtdx.TFSFPlaneSourceRegion
     fdtdx.TreeClass
+    fdtdx.TukeyWindow
     fdtdx.unfold_array
     fdtdx.unfold_detector_states
     fdtdx.unfold_fields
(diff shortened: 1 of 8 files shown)


bruxillensis:
@ymahlau I think this is ready for a review. I tried to be more thorough in testing.

l-berg:
Thank you for this PR and the thorough test cases.

My only gripe is with the naming scheme. Having *Profile and *WindowProfile classes that do different things might be confusing to other users. What do you think about using TemporalWindow / GaussianWindow / TukeyWindow instead?

Other than that, it looks fine to merge.

bruxillensis:
Agreed! Pushed the change

l-berg:
Great, thanks!

Issue #425, opened 2026-07-28
Symmetry bug(?) with spatially varying sources


ataras2:
First of all I want to say that I think this is a great package, thank you so much for developing and maintaining it!

I'm trying to simulate a larger device with x+y mirror symmetry. The input field is also symmetric but not spatially uniform.
A minimal example that recreates my issues is here: [minimal_source_bug.py](https://github.com/user-attachments/files/30464886/minimal_source_bug.py)

I place a `GaussianPlaneSource` in the center of my volume when running with symmetry on. After running the simulation, the source is still placed at the center of the quadrant when I expect it to be at the bottom left:
<img width="1236" height="413" alt="Image" src="https://github.com/user-attachments/assets/5c2d41dd-d6b6-45eb-8d71-53e32b1bedf2" />

Is there something I'm missing here? Have also tried moving the source but it doesn't seem to move.

ymahlau:
@ataras2 thanks for reporting this, symmetry is a new feature that we have not tested and documented well yet. I will make a PR to fix the bug soon :)

bruxillensis:
Confirmed. I tested this feature (originally) with ModeSource & UniformPlaneSource. ModeSources get oriented appropriately, Gaussians weren't tested since uniform sources are invariant to a placement shift.

PR #430, opened 2026-07-29
bug fix: spatially varying plane sources and PMC walls under mirror symme…

bruxillensis:
Fix for #425. I hope its alright I tackled this @ymahlau, I felt responsible :).

Four independent bugs made a symmetry-reduced simulation wrong for any source with spatial variation along a symmetry axis. All were invisible to the existing symmetry tests, which only use UniformPlaneSource with normalize_by_energy=False and direction="+" -- a combination blind to every one of them.

1. Profile center re-centered on the reduced quadrant. TFSFPlaneSource._get_center derived the transverse center from the object's clipped extent, so a Gaussian spot moved to the middle of the kept half instead of staying on the symmetry plane. The reduction discarded the pre-clip extent, so the source could not know it had been cut; reduce_resolved_slices now reports the per-axis low-side clip and place_objects records it as SimulationObject._symmetry_clip_low.

2. Normalization summed over the clipped plane. normalize_by_energy and the Gaussian profile's own sum normalization both divided by a quarter-plane sum, inflating the injected amplitude by 2**(n_axes/2) -- exactly 2x for two symmetric axes.

3. Profile mirrored about `center` along the vertical axis. v_basis = cross(wave_vector, u_basis) has a sign that depends on the propagation axis and direction, so the sampled profile was reflected for e.g. direction="-" along z (and, oppositely, direction="+" along y). Predates the symmetry feature by ~14 months and is invisible for any profile symmetric about its own center; it also blocked fix 1, since a correctly off-center profile got flipped out of the array.

4. The PMC symmetry wall over-constrained the field. Along a symmetry axis the six components split by Yee offset: tangential E and normal H sit at offset 0, normal E and tangential H at offset 1/2. The mirror plane lands on the reduced domain's low cell edge, so the tangential-H node a magnetic mirror must zero is at local -1 -- outside the array, where zero-padding already supplies it. Inserting a PerfectMagneticConductor instead zeroed tangential H at local 0, a full cell inside the domain. PEC still needs its wall (tangential E has no node on the plane). make_symmetry_walls now builds walls only for electric (-1) planes.

Validation. A symmetry-reduced run is compared cell-for-cell against the half of an equivalent full-domain run that it is supposed to represent (no unfolding involved). The figure quoted is the maximum absolute deviation of E over that half, divided by the peak |E| of the full-domain run -- so 0.0019 means the worst cell in the reduced domain is off by 0.19% of the full-domain field peak. Source is a Gaussian beam except where a dipole group is named. Values are before -> after this change:

  PEC on x/y/z (single axis)         0.0019 -> 0.0019   (unchanged)
  PMC on x/y/z (single axis)         0.2768 -> 0.0019
  PEC+PMC pairs (4 orientations)     0.2753 -> 0.0019
  PMC on x/y (mirror dipole pair)    0.6819 -> 0.0002
  PMC+PMC     (mirror dipole quad)   1.8677 -> 0.0003

PEC-only axes are untouched by this commit; they are listed for contrast, since they were already correct and are what made the PMC error stand out. The residual ~0.2% on the beam cases is dominated by PML reflection and beam truncation, not by the symmetry plane -- the dipole cases, which have neither, land at 0.02-0.03%.

The PMC fix converges under grid refinement -- 0.00142, 0.00053, 0.00032 at 200, 100 and 50 nm spacing -- so it is a consistent discretization rather than a cancellation that happens to work at one resolution.

With all four applied the injected reduced profile equals the kept half of the full-domain profile to float32 roundoff (~1e-6 relative).

**Known remaining limitation, not addressed here:** unfold_fields applies one array flip to all three components, but the correct reconstruction map differs by Yee staggering class. Measured over the PML-free interior, three components per axis reconstruct to ~0.1% of the full-domain peak and the other three to 2-12%, and which class is exact flips between axes. The reduced simulation itself is correct; only the unfolded reconstruction is affected.

**Found issue, not addressed here:** GaussianPlaneSource profile is transposed for propagation along y. GaussianPlaneSource._get_amplitude_raw builds its profile as (horizontal_axis, vertical_axis) and then inserts the singleton propagation axis. For propagation along y, get_oriented_transverse_axes(1) returns (2, 0) — horizontal = z, vertical = x — which is the only axis ordering where horizontal > vertical. The resulting array is therefore (z, 1, x) where the field slice expects (x, 1, z).

Consequences:
- Non-square transverse plane → hard failure. TypeError: div got incompatible shapes for broadcasting: (3, 12, 1, 16), (1, 16, 1, 12) raised from the impedance division in LinearlyPolarizedPlaneSource.apply.
- Square transverse plane → passes, because the built-in profile is circular with equal radii on both axes, so the transpose is numerically a no-op.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mirror-symmetry handling for reduced simulation domains.
  * Corrected plane-source orientation, centering, and amplitude normalization under symmetry reduction.
  * Magnetic symmetry boundaries now rely on natural domain padding instead of creating explicit walls.
  * Preserved object grid-shape information after symmetry clipping.

* **Tests**
  * Added coverage validating reduced-domain source profiles, amplitude scaling, symmetry clipping, and boundary behavior against full-domain simulations.

### DIFF
--- a/src/fdtdx/fdtd/initialization.py
+++ b/src/fdtdx/fdtd/initialization.py
@@ -162,8 +162,9 @@ def place_objects(
     # non-symmetric path is unchanged.
     dropped_names: set[str] = set()
     reduced_volume_shape = None
+    symmetry_clip_low: dict[str, tuple[int, int, int]] = {}
     if config.has_symmetry:
-        resolved_slices, dropped_names, reduced_volume_shape = reduce_resolved_slices(
+        resolved_slices, dropped_names, reduced_volume_shape, symmetry_clip_low = reduce_resolved_slices(
             resolved_slices=resolved_slices,
             object_map=object_map,
             config=config,
@@ -201,25 +202,26 @@ def place_objects(
         obj = object_map[name]
         assert key is not None
         key, subkey = jax.random.split(key)
-        placed_objects.append(
-            obj.place_on_grid(
-                grid_slice_tuple=slice_tuple,
-                config=config,
-                key=subkey,
-            )
+        placed = obj.place_on_grid(
+            grid_slice_tuple=slice_tuple,
+            config=config,
+            key=subkey,
         )
+        if name in symmetry_clip_low:
+            placed = placed.aset("_symmetry_clip_low", symmetry_clip_low[name])
+        placed_objects.append(placed)
 
     # Step 7: Place volume first (index 0)
     assert key is not None
     key, subkey = jax.random.split(key)
-    placed_objects.insert(
-        0,
-        volume_obj.place_on_grid(
-            grid_slice_tuple=resolved_slices[volume_obj.name],
-            config=config,
-            key=subkey,
-        ),
+    placed_volume = volume_obj.place_on_grid(
+        grid_slice_tuple=resolved_slices[volume_obj.name],
+        config=config,
+        key=subkey,
     )
+    if volume_obj.name in symmetry_clip_low:
+        placed_volume = placed_volume.aset("_symmetry_clip_low", symmetry_clip_low[volume_obj.name])
+    placed_objects.insert(0, placed_volume)
 
     # Step 8: Insert the PEC/PMC symmetry walls and forward the per-axis condition to mode
     # sources/detectors, then warn that the simulation now runs on the reduced domain.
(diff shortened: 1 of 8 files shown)


ymahlau:
@bruxillensis, my AI agent also did a fix over night, see #431 :D  I guess we duplicated work here a little. I would tend towards going forward with my PR because it also fixes ModeSource and works with non-square profiles

PR #441, opened 2026-08-28
fix(sources): interpolate point dipoles on Yee grids

Zhenhao-Xie:
## What Changed

- Added optional continuous-position interpolation for `PointDipoleSource`.
- Distributes electric and magnetic dipoles onto neighboring component-specific Yee samples.
- Supports rectilinear grids, local material sampling, dispersive permittivity, JIT, and gradients.
- Preserves the existing single-cell behavior by default.

## Why

Point dipoles were previously snapped to a single Yee-grid sample, introducing position-dependent discretization errors. Interpolation represents off-grid dipole positions more accurately while remaining compatible with existing simulations.

## Summary by CodeRabbit

* **New Features**
  * Added optional interpolation for point dipole sources across neighboring Yee-grid samples.
  * Supports interpolated electric and magnetic dipoles on uniform and non-uniform grids.
  * Preserves existing single-cell behavior by default.

* **Bug Fixes**
  * Improved material sampling for interpolated dipoles, including dispersive materials and material gradients.
  * Added validation for invalid interpolation settings and unsupported grid shapes.

### DIFF
--- a/src/fdtdx/objects/sources/dipole.py
+++ b/src/fdtdx/objects/sources/dipole.py
@@ -266,8 +266,6 @@ class PointDipoleSource(Source):
         tuple[SliceTuple3D, SliceTuple3D, SliceTuple3D],
         tuple[jax.Array, jax.Array, jax.Array],
     ]:
-        if isinstance(self._source_support_slices, Null) or isinstance(self._source_weights, Null):
-            return self._single_cell_support()
         return self._source_support_slices, self._source_weights
 
     @property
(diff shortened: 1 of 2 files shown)


ymahlau:
@Zhenhao-Xie I agree that this is an issue in the current implementation. However, I don't believe that this PR goes the best way towards fixing it. 
- In non-uniform grids, the volume weights are off and an update between a wide an narrow cell would probably be wrong
- +250 lines in dipole.py is more than the feature needs
- _source_support_slices should be frozen_private_field
- and many minor things that seem a bit off. 

I will close this PR for now, because we are currently busy with the re-write in pytorch. This would change many things anyways. If you really want this feature to be included now you can attempt to fix the points above, but I think it would not be worth the effort at the moment.

PR #446, opened 2026-09-14
automating the update mechanism for notebook build

renaissancenerd:
## Summary by CodeRabbit

- **Documentation**
  - Documentation navigation now automatically includes all notebooks in the Quickstart, Physics, Components, and Advanced sections.
  - Newly added notebooks in these sections can appear without updating explicit file lists.
  - Generated gallery card sections and related guidance have been removed from the tutorial pages.

- **Bug Fixes**
  - Notebook synchronization now preserves existing notebooks instead of removing them during updates.
  - Gallery cards are no longer regenerated as part of notebook synchronization.


ymahlau:
@renaissancenerd could you please describe how the process of adding a new notebook would look like now? I feel like with this solution is an overengineered AI-generated solution. There are probably more changes made here than necessary, right? Please only change the one thing of preventing the errors and keep visuals and process of adding notebooks the same if possible.

renaissancenerd:
> @renaissancenerd could you please describe how the process of adding a new notebook would look like now? I feel like with this solution is an overengineered AI-generated solution. There are probably more changes made here than necessary, right? Please only change the one thing of preventing the errors and keep visuals and process of adding notebooks the same if possible.

Yes I agree. I am currently simplifying it.

renaissancenerd:
@ymahlau Each `toctree` now uses `:glob:` with a single wildcard instead of a manual file list so that everything matches in each folder, sorted alphabetically (so order is preserved automatically).

ymahlau:
Very smart solution, awesome!

PR #450, opened 2026-09-21
fix(fdtd): reversible adjoint runs one reverse step past the start of the simulation

ThisSentenceIsALie:
# Fix: reversible adjoint runs one reverse step past the start of the simulation

## Summary

`reversible_fdtd`'s reverse loop executes one body call too many. The extra call reconstructs the
field at `t = -1` and pulls back a forward step that never happened. With an exact recorder this
is a silent ~1e-7 perturbation of `d(loss)/d(inv_permittivities)`; with a compressing recorder
(`LinearReconstructEveryK`) it turns most of that gradient into `NaN`.

## Cause

`body_fn` steps the reconstruction *back* before taking the VJP:

```python
def body_fn(sr_tuple):
    state, cot = sr_tuple
    state = backward(state=state, ...)        # time_step k -> k-1
    _, update_vjp = jax.vjp(forward_single_args_wrapper, state[0], ...)
    cot = update_vjp(cot)
    return state, cot
```

so entering the body at `time_step = k` back-propagates the forward step that *produced* state
`k`, i.e. forward step `k - 1`. The forward pass runs steps `0 .. time_steps_total - 1`, so the
last step needing a VJP is entered at `time_step = 1`, and the loop must stop on reaching `0`.
The guard is

```python
return time_step >= start_time_step      # start_time_step = 0
```

which lets the body run once more, entering at `0` and reconstructing `-1`.

At `t = -1` the recorder is asked for an interface it never stored. `Recorder(modules=[])` clamps
the out-of-range read to a neighbouring frame, which is close enough to the true `t ~ 0` interface
that the error stays around 1e-7. `LinearReconstructEveryK.decompress` instead interpolates
between two saved frames:

```python
interp_factor = (time_idx - prev_save_time) / (next_save_time - prev_save_time)
```

and at the degenerate index the denominator collapses, giving `0/0 -> NaN`. NaN is absorbing, so
the whole accumulated parameter cotangent is destroyed however small the term should have been.

## Fix

One character, `>=` -> `>`, plus a docstring recording why the body's decrement-before-VJP
ordering forces it.

## Evidence

Measured on a 731 x 50 x 46 cell run, 23996 steps, float32, comparing the reverse sweep step by
step against its own reconstructed primal and adjoint arrays:

* the first non-finite value appears at exactly one step, reverse step `nt + 1`, forward
  `time_step = -1`;
* the non-finite leaves are `primal_E`, `primal_H`, `cot_inv_permittivity`,
  `cot_inv_permeability`. Every adjoint *field* (`cot_E`, `cot_H`, `cot_psi_E`, `cot_psi_H`)
  stays finite, so this is the reconstructed primal poisoning the parameter cotangent, not a
  diverging adjoint;
* `|cot_inv_permittivity|` is 3.445e4 one chunk before the final step and 3.449e4 at it, i.e. the
  gradient is complete and correct and then NaN-stamped at the finish line;
* stopping the loop at `t = 0` makes the gradient finite and leaves the rest unchanged to 0.04%;
* with an exact recorder and the same bound change, `||g||` moves 19657.275 -> 19657.287
  (relative 6.0e-07), which is the size of the spurious contribution.

## Why the tests did not catch it

Every gradient test in `tests/simulation/fdtd/test_fdtd.py` builds a scene with `"periodic"` on all
six faces (`_build_scene`, `_build_slab_scene`, `_build_magnetic_lossy_scene`). A periodic scene
has no PML objects, so `input_shape_dtypes` is empty, nothing is recorded, and the recorder's
reconstruction is never called -- the machinery the reversible adjoint depends on is inert in every
test that differentiates. `LinearReconstructEveryK` is otherwise exercised only as a standalone
container in `tests/unit/interfaces`, never through a gradient.

The added test `tests/simulation/fdtd/test_reversible_recorder_gradient.py` puts a PML on all six
faces so the interface record is real, then takes a gradient with `LinearReconstructEveryK(k=8)`.
It fails on `main` (75.6% of the gradient non-finite) and passes with the fix.

## Validation

* new test on unpatched `main`: fails, `assert 0.755859375 == 0.0`
* new test with the fix: passes
* `tests/simulation/fdtd/test_fdtd.py tests/unit/fdtd tests/unit/interfaces` plus the new file
  with the fix applied: **456 passed**

One thing checked and deliberately *not* claimed: on a PML scene the reversible and checkpointed
gradients differ by 12x in whole-array norm, but agree to a ratio of 1.000001 outside the PML.
That is the documented behaviour of the reversible method (it replays the recorded interface
rather than reconstructing `d/d(eps)` inside the absorber), not a second defect, and the added
test does not assert on it.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed reversible gradient calculations to stop at the correct simulation timestep, preventing invalid reverse reconstruction and non-finite gradients.

* **Tests**
  * Added coverage for gradients through compressed recorder data in reversible and checkpointed FDTD simulations.
  * Verified that recorded simulations produce finite, nonzero parameter gradients.


ymahlau:
Thanks for catching this bug! Changes look good to me!

